Issue #25, opened 2015-02-11
Create config file


rgerstenberger:
I think we need a place to store configuration like default included mixins/modules and variables. I'm not sure if there's an npm convention we want to follow or pave our own path or something else.

/cc @alexlande @colinmegill @aykayen

alexlande:
@rgerstenberger: Can we close this given the addition of the radium-bootstrap config? Feels like this doesn't fit into Radium core.

rgerstenberger:
Yeah, until we have enough modules to warrant different builds this is a non-issue.

PR #26, opened 2015-02-11
Feature: Extend React.createClass for auto inline styles

rgerstenberger:
This adds the ability to require in `react-create-class-extended.js` to have radium auto generate styles and add them to props.  This technique is similar to the one in `react-style` but gives us more control over the current component and sub-elements/components https://github.com/js-next/react-style/blob/master/lib/ReactElementExtended.js.

usage:
```
var React = require('react');
// Replaces `React.createClass` with a version that calls original with a wrapped render function
var ReactCreateClassExtended = require('../../modules/react-create-class-extended.js');

var AutoButton = React.createClass({
  // `radiumStyles` is just a convention, the name is stored in a var and we can change it anytime
  radiumStyles: function () {
    return {
      standard: {
        fontSize: 16,
        color: "#fff"
      },
      refs: {
        my_ref: {
          color: "#aaa"
       }
    }
  },
  render: function () {
    return (
     // Top level element gets `styles=this.buildStyles(this.getStyles())`and 
     // `this.getBrowserStateEvents(/* compiled styles* /)` applied automatically
     <button>
        // ref that matches a ref key gets those styles applied
        <span ref="my_ref">{this.props.children}</span>
      </button>
    );
  }
});
```

Notes:
* Only runs if required in. 
* Only runs on components that adds `radiumStyles` to `createClass` (arbitrary name but good to standardize)
* Auto compiles `radiumStyles` and adds them to `props` on top level element returned from render
* *radiumStyles* can be either an object or a function
* Changes browser state so that user event handlers are now wrapped with radium event handlers
  - There was a recursive bug where the radium handler would also be the prop handler and form an infinite loop (pretty sure this exists outside of this branch)
* Added support for `refs` with contained element styles
  - this logic could easily be expanded to handle contained elements of all sorts
  - AKA style all `<li>`s returned from render
  - AKA `nth-child` selectors
  - AKA literally anything you can think of to handle contained elements e.g. all elements that contain `"dog"`
* Changed both mixin internals so that we can expose only the necessary pieces and keep other functions private.
* `getBrowserStateEvents` now takes styles object so it only adds handlers based on whitelisted states

Future plans:
* Add support for `ref` event states - tried and failed but I'm sure there's a way
* Reorganize files and functions 
  - Calling something as a mixin from radium itself is odd
  - I think we can abstract and move pieces around to make them more testable and maintainable
* I think we need a config file: https://github.com/FormidableLabs/radium/issues/25
* Use this to it's full potential!!!!! I think this opens the doors for a lot more complicated features
  - Not to say we can't do said complicated features with other approaches...

/cc @alexlande, @colinmegill, @aykayen, @eastridge, @kenwheeler, @ianobermiller


alexlande:
@rgerstenberger: Going to close this one if that's cool with you. Given the amount of API changes since this I think you're going to want a fresh PR when you revisit this.

PR #29, opened 2015-02-18
Style component

rgerstenberger:
This adds a style component that can be used with the following syntax:

`<Style selector="h2" styles={myStyles} />`

This component is very slow. Considering the main use case is for user generated content, I suggest we attempt to speed it up. Possibly with something like `insertRule` http://davidwalsh.name/add-rules-stylesheets. Really, all uses of this component should update a global stylesheet.

Also, I'm not happy with the API. If the main use case is unknown html on the page then there's probably more than a few selectors needed. I considered taking in an object like: 
```
{
 ".myClass": { background: "#fff"},
 ".otherClass: { background: "#000"}
}
```

But that logic could be very slow and the objects being passed in will be hard to validate. It's also might confuse the other radium style objects, so I opted to keep them similar. 

Lastly, we should probably be reusing the existing logic for building styles so that someone could pass in your average radium style object. 

This PR is not ready for primetime but I wanted to submit it to get clarity on which direction to pursue.

/cc @alexlande, @aykayen, @colinmegill

### DIFF
--- a/modules/components/style.js
+++ b/modules/components/style.js
@@ -1,0 +1,45 @@
+var React = require('react');
+var hyphenateStyleName = require('react/lib/hyphenateStyleName');
+var forEach = require('lodash.foreach');
+var reduce = require('lodash.reduce');
+
+function buildCssString (selector, rules) {
+  // TODO: Better error checking for rules object
+  if (!selector || !rules) {
+    return;
+  }
+
+  var cssString = selector + "{";
+
+  // Turn rules into css properties
+  cssString += reduce(rules, function (s, val, styleName) {
+    return s + hyphenateStyleName(styleName) + ":" + val + ";"
+  }, "");
+
+  // Close selector
+  cssString += "}";
+
+  return cssString;
+}
+
+var Style = React.createClass({
+  render: function () {
+    if (!this.props.selector) {
+      throw new Error("Style tag requires a 'selector' property");
+    }
+
+    // Build styles from standard
+    var styles = buildCssString(this.props.selector, this.props.standard)
+
+    forEach(this.props.states, function (rules, state) {
+      styles += buildCssString(ownerSelector + ":" + state, rules);
+    });
+
+    return (
+      <style dangerouslySetInnerHTML={{__html: styles}}>
+      </style>
+    );
+  }
+});
+
+module.exports = Style;
--- a/modules/components/style.js
+++ b/modules/components/style.js
@@ -24,14 +24,18 @@ function buildCssString (selector, rules) {
 
 var Style = React.createClass({
   render: function () {
-    if (!this.props.selector) {
-      throw new Error("Style tag requires a 'selector' property");
+    if (!this.props.selector || typeof this.props.selector !== "string") {
+      throw new Error("Style tag requires a 'selector' string");
+    }
+
+    if (!this.props.styles || typeof this.props.styles !== "object") {
+      throw new Error("Style tag requires a 'styles' object");
     }
 
     // Build styles from standard
-    var styles = buildCssString(this.props.selector, this.props.standard)
+    var styles = buildCssString(this.props.selector, this.props.styles.standard)
 
-    forEach(this.props.states, function (rules, state) {
+    forEach(this.props.styles.states, function (rules, state) {
       styles += buildCssString(ownerSelector + ":" + state, rules);
     });
 
--- a/modules/components/style.jsx
+++ b/modules/components/style.jsx
@@ -24,25 +24,24 @@ function buildCssString (selector, rules) {
 
 var Style = React.createClass({
   render: function () {
-    if (!this.props.selector || typeof this.props.selector !== "string") {
-      throw new Error("Style tag requires a 'selector' string");
+    var styles;
+
+    // Throw error if someone passes in valid children that isn't an object
+    if (this.props.children && typeof this.props.children !== "object") {
+      throw new Error("Style tag requires an object for 'props.children'");
     }
 
-    if (!this.props.styles || typeof this.props.styles !== "object") {
-      throw new Error("Style tag requires a 'styles' object");
-    }
-
-    // Build styles from standard
-    var styles = buildCssString(this.props.selector, this.props.styles.standard)
-
-    forEach(this.props.styles.states, function (rules, state) {
-      styles += buildCssString(ownerSelector + ":" + state, rules);
+    forEach(this.props.children, function (rules, selector) {
+      styles = buildCssString(selector, rules);
     });
 
-    return (
-      <style dangerouslySetInnerHTML={{__html: styles}}>
-      </style>
-    );
+    if (styles) {
+      return (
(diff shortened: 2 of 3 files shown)


alexlande:
> This component is very slow. Considering the main use case is for user generated content, I suggest we attempt to speed it up.

Could you elaborate on which portions are slow? Based on the `insertRule` + single stylesheet recommendation, I'm assuming you're talking about the render/paint cost of multiple separate `style` elements. Would like to not deal with that at the moment, with the understanding that implementing a single global style tag is on the roadmap.

Re: single vs multiple selectors-- we *need* multiple selector support.

> objects being passed in will be hard to validate

I don't follow, could you elaborate?

> Lastly, we should probably be reusing the existing logic for building styles so that someone could pass in your average radium style object.

I think keeping the `Style` API dumb (not baking in Radium stylesheet resolution) is fine. Users can do `buildStyles` in the parent component and pass the correct styles into the `Style` component.

modules/components/style.js
```diff
@@ -0,0 +1,49 @@
...
+
+    // Build styles from standard
+    var styles = buildCssString(this.props.selector, this.props.styles.standard)
+
+    forEach(this.props.styles.states, function (rules, state) {
+      styles += buildCssString(ownerSelector + ":" + state, rules);
+    });
+
+    return (
+      <style dangerouslySetInnerHTML={{__html: styles}}>
```
alexlande: Nit: I think you can write this as `<style dangerouslySetInnerHTML={{__html: styles}} />`

rgerstenberger: You are correct sir!

rgerstenberger:
Yes, slow because multiple individual style tags are slow. Ok we'll push off optimizing this for now :). 

>Re: single vs multiple selectors-- we need multiple selector support.

Agree, I'm not sure what a clean api for this looks like. Perhaps we take a different route and use `this.props.children` something like this?

```
// Not sure how this differs from using a style tag directly
<Style>
  h1 {
   background-color: "#ccc"
  }
</Style>

// If we want to be flexible, we could also handle an object
<Style>{
  h1: {
   backgroundColor: "#ccc"
  }
}</Style>
```

Back to multiple selectors, in order to accept objects of any sort we need more logic and we need to decide how we'll handle pseudo selectors. I guess you could have `{ "h1:hover": { styles } }`?

This was more stream of thought than expected.
1. Should we go with `this.props.children` vs specific props?
2. If we go with children should we accept string and objects?

cc @alexlande

alexlande:
I don't know if `this.props.children` is an option. It's prohibitively difficult to write CSS in JSX. React escapes everything, and chokes on the `{`. You have to use multi-line strings for it.

Re: pseudo-selectors, yeah, I think for this use case just writing them as `h1:hover` makes the most sense.

I actually have no idea if you can use a plain object as a component child. I would expect no, but not really sure. Not clear what the benefit over that syntax over just a single object with selectors and styles would be though.

rgerstenberger:
You are correct, CSS strings are tough. I refactored to use `this.props.children` but only accept objects or falsy values that don't render the style tag. It works as expected with the new syntax being `<Style>{ { selector: rules } }</Style>`.  Also added an example. Please re-review @alexlande

modules/components/style.jsx
```diff
@@ -0,0 +1,48 @@
...
+  render: function () {
+    var styles;
+
+    // Throw error if someone passes in valid children that isn't an object
+    if (this.props.children && typeof this.props.children !== "object") {
+      throw new Error("Style tag requires an object for 'props.children'");
+    }
+
+    forEach(this.props.children, function (rules, selector) {
+      styles = buildCssString(selector, rules);
```
alexlande: Do we want a reduce here? This replaces `styles` with each iteration, no? Doesn't support multiple rulesets.

ryan-roemer: @rgerstenberger @alexlande -- Something to consider for the whole project: React already requires es5 shims / shams (Function.prototype.bind stuff), so if you want to slim down the deps, you can switch to _just_ things in lodash that aren't ES5 (like `merge`).

alexlande: Good call :+1:

modules/components/style.jsx
```diff
@@ -0,0 +1,48 @@
...
+    });
+
+    if (styles) {
+      return (
+        <style dangerouslySetInnerHTML={{__html: styles}} />
+      );
+    } else {
+      return false;
```
alexlande: Nit: returning `null` is more idiomatic than `false` in this context.

rgerstenberger: Is it? I thought render wanted return false if there's nothing to show? Guess we're both right:

>You can also return null or false to indicate that you don't want anything rendered. Behind the scenes, React renders a <noscript> tag to work with our current diffing algorithm. When returning null or false, this.getDOMNode() will return null.

I'll change it to `null`

modules/components/style.jsx
rgerstenberger: Yes reduce. Will change

rgerstenberger:
@alexlande Updated, please re-review

modules/components/style.jsx
```diff
@@ -0,0 +1,45 @@
+var React = require('react');
+var hyphenateStyleName = require('react/lib/hyphenateStyleName');
+var reduce = require('lodash.reduce');
```
ryan-roemer: Per discussion with @alexlande -- can you change to native `ARRAY.reduce`?

@alexlande @rgerstenberger -- OR, feel free to ticket out and do globally...

alexlande: Yeah, I think it makes sense to do it in one big pass, will open a ticket.

modules/components/style.jsx
```diff
@@ -0,0 +1,45 @@
...
+
+function buildCssString (selector, rules) {
+  // TODO: Better error checking for rules object
+  if (!selector || !rules) {
+    return;
+  }
+
```
ryan-roemer: I'd actually suggest:

```js
return selector + "{" + rules.map(function (val, styleName) {
  return hyphenateStyleName(styleName) + ":" + val + ";";
}).join("") + "}";
```

alexlande:
Works great! :shipit:

modules/components/style.jsx
alexlande: Very clean :+1:

rgerstenberger: Definitely an improvement

rgerstenberger: Except it's an object, not an array. Is there an alternative to:
```
return selector + "{" + Object.keys(rules).map(function (key) {
  return hyphenateStyleName(key) + ":" + rules[key] + ";";
}).join("") + "}";
```

rgerstenberger: Updated to use my previous comment's code. @ryan-roemer do you approve?

ryan-roemer: totally!

And for this project, I'm just making suggestions, don't wait up for me ;)

Issue #34, opened 2015-02-19
Consider removing `standard` property in Radium style object


alexlande:
It may be cleaner to just keep styles at the top level, and not require a `standard` property.

eastridge:
As a noob I like that idea a lot, as it was my first instinct. Not sure how that plays well with everything else you need in the definition objects.

PR #35, opened 2015-02-19
Feature: document the thing

### LINKED ISSUE #36: Write guides

alexlande:
This adds three usage guides:

- Overview (basic usage, including standard styles, modifiers, and browser states)
- Media queries
- Computed styles

Resolves #36 

/cc @colinmegill @eastridge @rgerstenberger @aykayen @paulathevalley

### DIFF
--- a/docs/guides/overview.md
+++ b/docs/guides/overview.md
@@ -184,3 +184,5 @@ Then, add Radium browser state event listeners to your component:
 ```
 
 The `getBrowserStateEvents()` method returns a hash of events that Radium uses to set browser state on a component.
+
+For advanced usage, including media queries and computed styles, see our [guides](README.md).
--- a/docs/guides/overview.md
+++ b/docs/guides/overview.md
@@ -1,0 +1,186 @@
+# Using Radium
+
+Radium is a toolset for easily writing React component styles. It resolves modifiers, browser states, and media queries to apply the correct styles to your components, all without selectors, specificity, or source order dependence.
+
+## How do I do it, then?
+
+Let's create a fictional `<Button>` component. It will have a set of default styles, will adjust its appearance based on modifiers, and will include hover, focus, and active states.
+
+```js
+var Button = React.createClass({
+  render: function () {
+    return (
+      <button>
+        {this.props.children}
+      </button>
+    );
+  }
+});
+```
+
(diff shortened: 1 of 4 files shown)


docs/guides/overview.md
```diff
@@ -0,0 +1,188 @@
...
+// if block === true
+block: {
+  display: 'block'
+}
+```
+
+By default, modifiers map directly to a component's `props`:
+
+```js
+<Button
```
eastridge: It might make sense to open the section with this example and then explain how it works.

docs/guides/overview.md
```diff
@@ -0,0 +1,188 @@
...
+
+  render: function () { ... }
+});
+```
+
+`StyleResolverMixin` has a method called `buildStyles()` that converts a nested Radium style object into the styles that should be applied to your component based on its props and state. A basic Radium style object looks like this:
+
+```js
+{
+  standard: {
```
eastridge: It's not clear to me wether or not `standard` is a required thing, or wether I could put all of my styles on the root level of this object.

eastridge: Other thought: we could introduce "yes it's CSS in JS" earlier, then focus on just what the `StyleResolverMixin` is doing here.

alexlande: It's required, but over the course of writing this I started thinking styles should just be at the root instead: https://github.com/FormidableLabs/radium/issues/34

eastridge: Could further change this code block to: `var radStyles = ...` so that the example you show below is valid vs "where did `radStyles` come from, is it special?"

docs/guides/overview.md
```diff
@@ -0,0 +1,188 @@
...
+      display: 'block'
+    }
+  }
+}
+```
+
+Modifiers can reflect string or boolean values. If a modifier value is a string, you can represent different possible values as child objects with their own CSS properties:
+
+```js
+size: {
```
eastridge: Show the complete object, i.e. `{modifiers: {size:...`

alexlande:
Updated based on @eastridge's feedback and API changes since this was opened. Going to merge in now for the sake of keeping things moving. If anyone has any additional feedback or suggestions, please open an issue or submit a PR!

PR #39, opened 2015-02-19
Remove styles.standard pattern

### LINKED ISSUE #34: Consider removing `standard` property in Radium style object

alexlande:
This may have been useful at some point, but not anymore. Cleaner API for everybody!

Resolves #34 

/cc @eastridge @colinmegill @rgerstenberger @aykayen @paulathevalley


eastridge:
:clap:

Issue #31, opened 2015-02-19
Replace lodash with ES5 methods where possible


alexlande:
We should keep our dependencies as slim as possible. To that end, we should favor ES5 array and object methods (that React shims anyway) to their lodash equivalents where possible. This means you, `forEach`, `map`, and `reduce`!

PR #40, opened 2015-02-19
Chore: Clean up lodash usage

### LINKED ISSUE #30: Convert separate lodash packages to lodash path includes
Currently we're installing and importing separate `lodash` packages. It's easier to just add `lodash` as a dependency then require specific lodash modules by path. Instead of:

```
var merge = require('lodash.merge');
```

We can do:

```
var merge = require('lodash/object/merge');
```

alexlande:
This replaces our individual lodash packages with the normal `lodash` package, which supports requiring deep linked individual methods, which is very cool. Thanks @ryan-roemer for pointing that out.

Also replaced some instances of `forEach` with `for...in` loops where appropriate.

Resolves #30 and #31

/cc @ryan-roemer @eastridge @rgerstenberger @colinmegill

### DIFF
--- a/modules/mixins/style-resolver.js
+++ b/modules/mixins/style-resolver.js
@@ -1,5 +1,4 @@
-var forEach = require('lodash.foreach');
-var merge = require('lodash.merge');
+var merge = require('lodash/object/merge');
 
 var StyleResolverMixin = {
   _getStateStyles: function (states) {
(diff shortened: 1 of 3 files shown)


alexlande:
Modified slightly. This has made it clear that we need to make modifiers, mediaQueries, and states arrays so that they're evaluated in order, to be correct, and you know, good.

PR #42, opened 2015-02-20
Feature: Scoped style component

### LINKED ISSUE #32: Scoped style component
Splitting this into it's own issue from #18.

## Scoped style tags

For things like user-generated content, or places where we need to use pseudo-elements, nth-child, etc, we need `style` tags that are scoped to the component they're included in. This is another case where you can use regular CSS in the meantime. I think this would look like this:

```
    <div>
      <Style local={true} style={scopedStyles} />
      {this.getUserGeneratedContent()}
    </div>
}
```

This would return a style tag where any rules would be scoped to the parent component. Given that the parent div would have some unique identifier ("we'll say `#react-radium-id-foo-123`):

```
<style>
#react-radium-id-foo-123 a {
  color: blue;
}

#react-radium-id-foo-123 a:hover {
  text-decoration: underline;
}
</style>
```

rgerstenberger:
This closes #32.

Adds an optional prop to `Style` component `scopeSelector` that allows for all passed in styles to be scoped to that selector.

/cc @alexlande, @aykayen, @colinmegill

### DIFF
--- a/examples/app.jsx
+++ b/examples/app.jsx
@@ -37,6 +37,15 @@ var App = React.createClass({
         <p>
           <ComputedWell>Click me!</ComputedWell>
         </p>
+
+        <p className="scoping-class">
+          <Style scopeSelector=".scoping-class">{{
+            span: {
+              fontFamily: "\"Lucida Console\", Monaco, monospace"
+            }
+          }}</Style>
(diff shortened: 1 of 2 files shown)


alexlande:
:shipit:

PR #44, opened 2015-02-22
Feature: Replace matchMedia store with context-based mixins

alexlande:
Finally read about `context` this morning and realized that it's a perfect fit for Radium media queries. With this update, there's no more `matchMedia` store and no need to pass media queries down as state or props. They're accessible to any descendants of the element that initializes them with the `MatchMediaBase` mixin.

/cc @colinmegill @eastridge @kenwheeler @colinmegill @aykayen @paulathevalley @ryan-roemer


package.json
```diff
@@ -19,15 +19,12 @@
     "examples": "webpack-dev-server --config examples/webpack.config.js --no-info --content-base examples/"
   },
   "license": "MIT",
-  "peerDependencies": {
-    "react": "0.12.x"
-  },
   "dependencies": {
+    "react": "0.12.x",
```
ryan-roemer: Can you give me a little more background into this being a real dependency? Do we need to switch to target any particular builder (e.g. webpack)?

alexlande: Wasn't related to build. These new mixins do `require('react')` directly, so it seemed like the correct thing to do. That said, `peerDependencies` is still sort of mysterious to me, so could be unnecessary?

ryan-roemer: Hmm... so it looks like Radium doesn't have a dist build yet. So here's my understanding (which may be very wrong) -- for things that you use internally and want added, add a `dependencies` entry, for things where you are the add-on, add a `peerDependencies` and `devDependencies` for your dev stuff. The reasoning is that your users are choosing React first, and then something like Radium as the add-on.

... but I'm not sure about how the packaging actually matters. Maybe one of us (or someone else at Formidable who's interested) see what the canonical / most typical answer is?

For a contrast, in `chai-jq`, I do a UMD wrapper: https://github.com/FormidableLabs/chai-jq/blob/master/chai-jq.js#L500-L528 but actually do not **not** have either `peerDependencies` or `dependencies` on Chai or jQuery, even though both are required for the package... https://github.com/FormidableLabs/chai-jq/blob/master/package.json

alexlande: Ok, just re-read the `peerDependencies` blog post here: http://blog.nodejs.org/2013/02/07/peer-dependencies/.

I was thinking of it in this context:

> Most plugins never actually depend on their host package, i.e. grunt plugins never do require("grunt"), so even if plugins did put down their host package as a dependency, the downloaded copy would never be used.

But that post specifies that `peerDependencies` should still be used even if your package has a direct dependency on the package in question:

> Even for plugins that do have such direct dependencies, probably due to the host package supplying utility APIs, specifying the dependency in the plugin's package.json would result in a dependency tree with multiple copies of the host package—not what you want.

Will switch back to `peerDependencies`. Good point though, need to set up dist for the non webpack/browserify crowd.

PR #55, opened 2015-02-28
Add standalone distribution

kenwheeler:
Adds webpack conf and npm task to build standalone distribution for use
in reduced test cases and demo code

cc @alexlande


webpack.config.js
```diff
@@ -0,0 +1,25 @@
+var path = require('path');
+var webpack = require('webpack');
+
+module.exports = {
+  cache: true,
+  entry: "./modules/index.js",
+  output: {
+    path: path.join(__dirname, '/dist'),
+    filename: 'Radium.js',
```
alexlande: Nit: can we lowercase the file name?

alexlande:
Thanks for this :ship:

alexlande:
Actually-- should we not bundle `React` in `dist`?

kenwheeler:
Yeah, great call. What do you think about lodash?

webpack.config.js
kenwheeler: Absolutely.

kenwheeler: Should the global var be lowercase too? 

```js
radium.MatchMediaItem
// vs
Radium.MatchMediaItem
```

alexlande: We've been using uppercase everywhere, but yeah, it probably should be lowercase.

alexlande:
I think we can leave lodash in, it's just a few methods.

package.json
```diff
@@ -16,6 +16,7 @@
     "Alex Lande"
   ],
   "scripts": {
+    "dist": "webpack --config ./webpack.config.js",
```
ryan-roemer: GLOBAL: Never prefix with `./`. Windows is compatible as long as you don't do that. (Internal `/` are fine).

ryan-roemer: NIT: You can also shorten this to `webpack` ;)

alexlande: Good call :) Thanks @ryan-roemer!

webpack.config.js
ryan-roemer: I vote lower case file name, uppercase class name. See, e.g.,

* https://code.jquery.com/jquery-2.1.3.min.js
* https://cdnjs.cloudflare.com/ajax/libs/backbone.js/1.1.2/backbone-min.js
* https://cdnjs.cloudflare.com/ajax/libs/react/0.12.2/react.min.js

alexlande: Works for me

ryan-roemer:
Also on the lower casing thing, be aware that in this specific PR from `dist/Radium.js` to `dist/radium.js` is likely going to mess things up on your mac's (which have a case insensitive file system even though it seems case sensitive). I'd recommend nuking all of `dist` first...

kenwheeler:
Ok, updated everything per discussion

alexlande:
Nice! :shipit:

kenwheeler:
For the record, that mouse means merge?

alexlande:
Yep :D

PR #62, opened 2015-03-02
Refactor style component

### LINKED ISSUE #57: Use React logic to serialize styles for style component.
To generate the innerHTML for each rule of the `<Style>` component, you could do something like this:

```js
require('react/lib/CSSPropertyOperations').createMarkupForStyles({
  width: 10,
  margin: 'auto'
});
// => 'width:10px;margin:auto'
```

alexlande:
This updates the `<Style>` component and gets it ready for release.

- Added `Style` documentation.
- Improved CSS string generation (thanks to a suggestion in #57)
- Changed `Style` syntax to be more idiomatic. Previously was using `this.props.children` for the passed in rules object and doing some error handling to make sure it was valid. This makes more sense than overloading `children` for a purpose other than the intended one.
- Added `propTypes`
- Added `Style` as an export from `index.js`

Resolves #57

/cc @rgerstenberger @colinmegill @kenwheeler @eastridge @ryan-roemer


docs/api/README.md
```diff
@@ -192,3 +192,46 @@ MatchMediaBase.init({
...
+```
+
+#### scopeSelector
+
+A string that any included selectors in `rules` will be appended to. Use to scope styles in the component to a particular element. A good use case might be to generate a unique ID for a component to scope any styles to the particular component that owns the `<Style>` component instance.
+
+```js
+<div class="TestClass">
+  <Style
+    scope=".TestClass"
```
eastridge: is `scope` or `scopeSelector` what should be used?

alexlande: Ah, `scopeSelector`. Thanks!

docs/api/README.md
```diff
@@ -192,3 +192,46 @@ MatchMediaBase.init({
 ## MatchMediaItem
 
 The match media item mixin applies media queries set in `MatchMediaBase` to a component. To use, add it as a mixin to any component that is a descendent of the component with `MatchMediaBase` that should include media query styles.
+
+## Style Component
+
+The `<Style>` component renders an HTML `<style>` tag containing a set of CSS rules. Using it, you can define an optional `scopeSelector` that all selectors in the resulting `<style>` element will include.
+
```
eastridge: Why would I use this versus just injecting a plain old `<style>`?

alexlande: Will add a note about it. (Basically, writing a `<style>` tag is really difficult in JSX. You have to write a multi-line string with all of your rules because JSX tries to turn everything into elements.)

ryan-roemer: I'm just jumping in here without doing any other back reading / homework, but is it an issue at all that `<Style>` is a totally valid HTML tag in any casing?

ryan-roemer: OK, did my back reading -- http://facebook.github.io/react/docs/jsx-in-depth.html#html-tags-vs.-react-components -- we're fine ;)

ryan-roemer:
@alexlande -- LGTM.

rgerstenberger:
@alexlande can we memorize the building of the CSS string?

PR #67, opened 2015-03-04
added minified webpack configuration & distribution

vincentriemer:


alexlande:
Thanks for this! Could you make two changes before I merge it in?

- [ ] Revert the changes to the two dist files (want to only build them when we tag a version to reduce repo churn)
- [ ] Update the minified webpack config to extend the base one with `_.assign()` to DRY things out a little bit.

vincentriemer:
Changes made, only thing different from what you suggested was using _.merge() instead of _.assign()

alexlande:
Perfect! Thanks a lot.

webpack.config.minified.js
```diff
@@ -0,0 +1,15 @@
+var webpack = require('webpack');
+var _ = require('lodash');
+
+module.exports = _.merge(require('./webpack.config.js'), {
```
ryan-roemer: Switch to

```js
module.exports = _.merge({}, require('./webpack.config.js'), {
```

If you end up using 2+ configurations in the same process (aka, what often happens in gulp-land...)

@alexlande -- Want to ticket that out?

alexlande: Ah, good call. Will do.

webpack.config.minified.js
```diff
@@ -0,0 +1,15 @@
...
+  output: {
+    filename: 'radium.min.js'
+  },
+  plugins: [
+    new webpack.optimize.UglifyJsPlugin({
```
ryan-roemer: Also suggest:

```js
new webpack.optimize.DedupePlugin()
```

now and in the future, consider writing "dev-only" code against a `production.env.NODE_ENV` check and then defining this:

```js
    new webpack.DefinePlugin({
      "process.env": {
        // Signal production mode for React JS and other libs.
        NODE_ENV: JSON.stringify("production")
    })
```

As that's more-or-less the react standard now...

PR #71, opened 2015-03-06
Fixed top-level-component sample code.

srph:
I got confused the first time I read the documentation. I thought the `createClass` had a slightly different API from what I've read the last time. And then, I rechecked, seems like there was a mistake.


alexlande:
Thanks! Who knows where I came up with that code example.

Issue #66, opened 2015-03-04
Active state should not be lost on mouseout if the mouse is held


ocharles:
In Chrome, if I click a button, hold the mouse, and then move the pointer outside the button, the button remains in the active state. However, using Radium, this behaviour is not respected, and I lose active state when the mouse leaves the button - regardless of the state of the mouse.

PR #81, opened 2015-03-09
Preserve active state on mouseout when the mouse is held

david-davidson:
Fixes https://github.com/FormidableLabs/radium/issues/66

This adds a helper module (`mixins/util/global-mouseup`) that sets up a single window-level `mouseup` listener. Components subscribe to (and unsubscribe from) it behind the scenes via `mixins/browser-state`, which now has a `handleGlobalMouseUp` method. Finally, `_handleMouseLeave` no longer sets the `active` state to false.

/cc @alexlande


alexlande:
Going to kick the tires on this one, but looks awesome so far.

modules/mixins/util/global-mouseup.js
```diff
@@ -0,0 +1,36 @@
...
+var _mouseUpListenerIsActive = false;
+
+var _handleMouseUp = function (ev) {
+  _listeningComponents.forEach(function (component) {
+    component.handleGlobalMouseUp(ev);
+  });
+};
+
+var subscribeToGlobalMouseUp = function (component) {
+  if (!includes(_listeningComponents, component)) {
```
ryan-roemer: I think you can avoid this dependency with normal `ARRAY.indexOf()`...

modules/mixins/browser-state.js
```diff
@@ -1,4 +1,17 @@
+var globalMouseUp = require("./util/global-mouseup");
```
ianobermiller: How about naming the module MouseUpListener and providing a simple method, subscribe, that takes a callback and returns an object with a `remove()` method. Then, BrowserStateMixin is not tightly coupled, and you don't have the magic of handleGlobalMouseUp being called across files.

modules/mixins/browser-state.js
```diff
@@ -59,6 +71,17 @@ var BrowserStateMixin = {
     });
   },
 
+  radiumGlobalMouseUp: function (ev) {
+    if (ev.toElement === this.getDOMNode()) {
```
alexlande: Given the desired behavior, is there any reason we couldn't remove this check *and* remove the mouseUp behavior in `browser-state.js`? https://github.com/FormidableLabs/radium/blob/master/modules/mixins/browser-state.js#L54-L60

Then the global mouseUp could handle both cases.

modules/mixins/util/global-mouseup.js
```diff
@@ -0,0 +1,38 @@
+var _listeningComponents = [];
+var _mouseUpListenerIsActive = false;
```
alexlande: What do we need `_mouseUpListenerIsActive` for?

alexlande: Scratch that, I get it.

david-davidson:
@alexlande @ianobermiller @ryan-roemer Thanks for the good suggestions!

modules/mixins/util/mouse-up-listener.js
```diff
@@ -0,0 +1,39 @@
+var _callbacks = [];
+var _mouseUpListenerIsActive = false;
+
+var _includes = function (arr, item) {
+  return arr.indexOf(item) !== -1;
```
alexlande: Let's drop this function and just do this check on line 15:

```js
if (arr.indexOf(item) === -1) {
  _callbacks.push(callback);
}
```

alexlande:
Awesome, thanks @david-davidson!

PR #84, opened 2015-03-10
Use babel-loader instead of jsx-loader

### LINKED ISSUE #83: Use babel-loader instead of jsx-loader
Any objections to using babel with all the ES6 goodness?

ianobermiller:
So we can have more ES6 happiness.

Fixes #83


alexlande:
I'm not opposed to this but would like to think on it for a bit. Right now, we use the JSX loader for the examples, but Radium doesn't need to be compiled.

If we start using Babel, we'll need to add compiled source for npm like react-router does, right? https://github.com/rackt/react-router/tree/master/build

ianobermiller:
Yeah you're right, and that is annoying. I'll give plain old pre ES5 JS a go instead for now.

ianobermiller:
We should be able to use a [prepublish](https://docs.npmjs.com/misc/scripts) script to do this with minimal work.

kenwheeler:
For babel based projects, I've been compiling to a .gitignored `lib` directory and pointing `main` at its index. For the time being I've been building and then `npm publish`'ing, which wraps up the current dir and sends it up, so then `lib` never makes it into version control, but I'm working on getting that step running in CI.

alexlande:
Yeah, `prepublish` seems like the right thing to do here. :+1:

alexlande:
Closing, covered in #144

PR #85, opened 2015-03-11
Remove redundant condition.

iclanzan:
The length of `_callbacks` is always greater than 0 at this point in the code.


alexlande:
Nice, thanks!

PR #86, opened 2015-03-12
Basic touch support

iest:
Not sure what you had in mind for touch support, but this seems to work well enough...

### DIFF
--- a/modules/mixins/browser-state.js
+++ b/modules/mixins/browser-state.js
@@ -23,6 +23,9 @@ var BrowserStateMixin = {
       onMouseEnter: this.radiumMouseEnter,
       onMouseLeave: this.radiumMouseLeave,
       onMouseDown: this.radiumMouseDown,
+      onTouchStart: this._handleMouseDown,
+      onTouchEnd: this._handleMouseUp,
+      onTouchCancel: this._handleMouseUp,
       onFocus: this.radiumFocus,
       onBlur: this.radiumBlur
     };


alexlande:
I'm not that well versed in touch events, but agree that this seems good enough for now. @kenwheeler?

alexlande:
That said, the implementation here is missing a few things + could use some adjustments:

- [ ] The `_handle` naming convention has been replaced with `radium`, so these should be updated to match that: `this._handleMouseUp` -> `this.radiumMouseUp`
- [ ] These will need their own `radiumTouchStart`/etc methods, rather than piggybacking on the existing mouseUp + mouseDown (these methods need to call their corresponding handlers in `this.props` if a user wants to use their own methods, see https://github.com/iest/radium/blob/master/modules/mixins/browser-state.js#L43)

Also, thanks for the PR!

david-davidson:
I wonder if `React.initializeTouchEvents(true)` would apply here? http://facebook.github.io/react/docs/events.html#touch-events

alexlande:
@david-davidson: Good call! We should use this: https://github.com/facebook/react/blob/4feed4fcc54dd81c60d668e4d3f38a88443b5c7f/src/event/EventPluginUtils.js#L204-L216

kenwheeler:
@david-davidson indeed.

iclanzan:
Unfortunately it is not as easy as binding the mouse event handlers to the touch events.

Think of the following scenario: You put your finger down on a button, triggering `touchStart` and setting an active state, then you drag the finger off the button and lift it. The `touchEnd` would not be triggered for the button and the active state would not be removed.

kenwheeler:
@iclanzan Wouldn't the `touchCancel` event pick that up?

iclanzan:
I wrote a simple [jsbin example](http://jsbin.com/zijarubobu), and by the looks of it `touchEnd` is being triggered so this might not be an issue after all.

Another problem can occur from the fact that on mobile devices mouse events are being triggered after touch events.

alexlande:
Re `touchEnd`: cool! Makes sense.

Re mouse + touch events: Is that the case, typically? I thought most mobile browsers only fire synthetic mouse events *if* no touch events are caught first.

If that is an issue, we could apply the touch listeners when `EventPluginUtils.useTouchEvents` is true and the mouse listeners when it's false.

iclanzan:
The synthetic mouse events are fired unless `event.preventDefault()` is called in the touch event handlers.

alexlande:
Right, that makes sense. Conditionally applying listeners based on `useTouchEvents` seems like the correct workaround though.

iclanzan:
What about devices that have both mouse and touch, like most Chromebooks?

alexlande:
Touch devices are the worst! :)

iclanzan:
I know, I’ve never been able to make them cooperate completely.

alexlande:
Going to close this, won't fit into the new API. Will look into what kind of touch handling we need shortly.

Issue #53, opened 2015-02-28
Aggregate style component for server-side rendering 


alexlande:
## Problem

Radium media queries and browser states don't work with server-side rendering. This hurts initial render speed (we lose some of the benefits of server-side rendering because the initial render is incorrect), and means that apps using Radium won't display correctly until/if the app is hydrated on the client.

Aside from looking bad, this is an accessibility concern, particularly for browser state events.

## Solution?

@rgerstenberger spent some time testing out variations of the `Style` component to handle this use case. In his tests, he handled `:hover`, `:focus`, and `:active` with style tags in each component, scoped to the parent component's react-id. The performance there was terrible, because the browser had to re-evaluate everything every time it hit a style tag.

The only good solution I can think of is to implement an aggregate `style` tag (code name: `StyleCatcher`) at the top of Radium apps. On server-side renders, we would take any media query + browser state styles and drop them there in a way similar to the existing `Style` component. I'm assuming this will have enough overhead that we won't want to do it on repeated renders on the client-side. If that's *not* the case, we could replace our existing media query + browser state behavior with the `StyleCatcher`.

## Challenges

- How do we now if it's a server side render or not (if we need to tell the difference)?
- How does the interaction between components and `StyleCatcher` work? When a component calls `buildStyles` it updates state on `StyleCatcher`? Probably.
- More things I'm forgetting?

alexlande:
I must be going crazy because my comments seem to be disappearing. Based on some productive Twitter conversation, going to aim to implement server-side media queries and browser states with individual style elements to start, then optimize to a `StyleCatcher` based on that.

lanceharper:
> How do we now if it's a server side render or not (if we need to tell the difference)?

Not sure if you're still exploring or already found this, but check out
https://github.com/facebook/react/blob/master/src/vendor/core/ExecutionEnvironment.js#L30

alexlande:
@lanceharper: That's perfect. Thanks!

alexlande:
Update on this. We've got the base `Style` component merged in. I'll be tagging a release later today with that. From here, we need to take care of the following to get server-side `Style` component rendering:

- [ ] Add media query API to `Style` (currently it only takes selectors and rules, no support for `@` rules)
- [ ] Figure out how the conditional `Style` components will be handled. Ideally, they should be rendered by `StyleResolverMixin` transparently if you're on the server. Not sure how feasible that is.
[...]

PR #91, opened 2015-03-16
Add media query API to `Style`

david-davidson:
This adds media query support to `Style`, which brings us one step closer to [server-side `Style` rendering](https://github.com/FormidableLabs/radium/issues/53#issuecomment-77470494).

To set up media queries, add a `mediaQueries` object to the `rules` array you pass to `Style`. Like `rules`, `mediaQueries` should be an array of objects with selectors as keys and style objects as values. You can use strings to define queries, as in
```
'(max-width: 500px)': [
  // styles
]
```
...or you can reuse previously `MatchMediaBase.init`ed queries, as in
```
MatchMediaBase.init({
  lg: '(min-width: 1200px)'
});

...

lg: [
  // styles
]
```
(The changes to `mixins/match-media-base` and `mixins/style-resolver` are to expose the `MatchMedia` queries via `MatchMediaItem`.)

/cc @alexlande

### DIFF
--- a/modules/components/style.js
+++ b/modules/components/style.js
@@ -49,27 +49,30 @@ var Style = React.createClass({
   },
 
   _buildMediaQueryString: function (mediaQueryObj) {
-    var mediaQueries = "";
+    var contextMediaQueries = this._getContextMediaQueries();
+    var mediaQueryString = "";
 
     Object.keys(mediaQueryObj).forEach(function (query) {
-      var completeQuery = this.contextMediaQueries[query] ?
-        this.contextMediaQueries[query] :
+      var completeQuery = contextMediaQueries[query] ?
+        contextMediaQueries[query] :
         query;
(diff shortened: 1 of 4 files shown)


modules/components/style.js
```diff
@@ -24,18 +27,58 @@ var Style = React.createClass({
...
     }
 
-    var styles = reduce(this.props.rules, function (s, item) {
-      var selector = Object.keys(item)[0];
-      var rules = item[selector];
-      var completeSelector = this.props.scopeSelector + " " + selector;
-
-      return s += buildCssString(completeSelector, rules);
-    }, "", this);
+    this._prepareContextMediaQueries();
```
alexlande: Would prefer to avoid caching additional data on `this`. Let's go for a more functional approach.

alexlande:
This is great, awesome work!

Issue #80, opened 2015-03-09
Eliminate getBrowserStateEvents and buildStyles


ianobermiller:
I've been thinking a lot about how we can simplify the API surface of Radium. As we've already seen from #76, it can be burdensome to remember to add getBrowserStateEvents and then wire up your own handlers. It is also not ideal to have to call buildStyles. Further, Radium only allows you to style a single element in the component (correct me if I'm wrong).

The only way to simplify this API is by somehow wrapping the render functions. There are a ton of ways to accomplish this, and they don't necessarily require monkeypatching createClass. I've been playing around with my own inline style helper, jokingly named Cesium: https://github.com/ianobermiller/nuclearmail/blob/master/src/js/Cesium.js. The main goals of Cesium are to:
- Be compatible with normal inline styles (you only include Cesium if you want interactivity or vendor prefixing)
- Be as simple and foolproof as possible to use

The syntax I'm currently using looks like this, simplified for the example. Note that the only differences between using normal inline styles are: 1) `Cesium.wrap` inside of `createClass`, and 2) `':hover'` prop with a style object value (this is basically a more concise version of `modifiers: { hover: { ... } }`

```js
var Button = React.createClass(Cesium.wrap({
  propTypes: {
    onClick: PropTypes.func,
  },
  _onClick() {
    this.props.onClick && this.props.onClick();
  },

  render(): any {
    return (
      <button type="button" style={styles.root} onClick={this._onClick}>
        {this.props.children}
      </button>
    );
  }
}));

var styles = {
  root: {
    background: Colors.gray1,

    ':hover': {
      background: Colors.gray1.darken(5),
    },

    ':active': {
      background: Colors.gray1.darken(10),
    },
  },
};
```

The main idea here is that you don't need `getBrowserStateEvents` and `buildStyles`. `Cesium.wrap` wraps the `render` method, walks the result of the original render, and updates styles/adds handlers as appropriate.

In the simple case of one interactive component, you don't need anything more than this. If you want to style multiple elements in a single component, you have to specify a unique `ref` or `key` for each one, to correlate the interactive state from the event handlers.

I'd love to see these ideas make their way into Radium. Thoughts?

alexlande:
Generally-- I like this a *lot*. Moving this direction would solve a lot of our issues, particularly our verbosity and the difficulty in applying browser states to multiple elements in a single component (without making them components themselves).

[...]

Issue #82, opened 2015-03-10
Eliminate modifiers


ianobermiller:
In my quest to simplify Radium, I propose we eliminate modifiers. Instead, if we implement something like #80, we can make the style object accept an array, and apply provided styles in order (ignoring non-object ones) [1]. Also, in this way, the component author can choose whether or not to include this.props.style. Finally, each state can specify :hover or :active states, and the merging would have to be somewhat intelligent about it [2]. I posit that this behavior is less magical than modifiers, and much more explicit & flexible.

[1]
```js
var Button = React.createClass(Cesium.wrap({
  propTypes: {
    use: PropTypes.oneOf(['default', 'special']),
    style: PropTypes.object,
  },

  render(): any {
    return (
      <button
        type="button"
        style={[
          styles.root,
          (this.props.use === 'default') && styles.default,
          (this.props.use === 'special') && styles.special,
          this.props.style
        ]} />
    );
  }
}));

var styles = {
  root: {
    border: 'none',
    borderRadius: '2px',
    cursor: 'pointer',
    lineHeight: '30px',
    margin: 0,
    padding: '1px 16px',
    verticalAlign: 'top',

    ':active': {
      padding: '2px 15px 0 17px',
    },
  },

  default: {
    background: Colors.gray1,
    color: Colors.black,

    ':hover': {
      background: Colors.gray1.darken(5),
    },
  },

  special: {
    background: Colors.accent,
    color: Colors.white,

    ':hover': {
      background: Colors.accent.darken(5),
    },
  },
};
```

[2]
```js
// Merge style objects. Special casing for props starting with ';'; the values
// should be objects, and are merged with others of the same name (instead of
// overwriting).
function _mergeStyles(...styles: Array<Object|boolean>): any {
  var styleProp = {};

  styles.forEach(style => {
    if (!style || typeof style !== 'object') {
      return;
    }

    var styleObj: Object = style; // flow :(

    Object.keys(styleObj).forEach(name => {
      if (name.indexOf(':') === 0) {
        styleProp[name] = styleProp[name] || {};
        Object.assign(styleProp[name], styleObj[name]);
      } else {
        styleProp[name] = styleObj[name];
      }
    });
  });

  return styleProp;
}
```

alexlande:
I tend to think that the convenience of handling modifiers is one of our better features. Your example is ok for a couple of modifiers (although still a bit verbose for such a common behavior, IMO):

```js
style={[
  styles.root,
[...]

PR #101, opened 2015-04-02
New wrap API

ianobermiller:
Basically an import of Cesium from
https://github.com/ianobermiller/nuclearmail/blob/e30abd3d3c06ff3fb0ed9e
db0c28b758b2cbd114/src/js/Cesium.js, ran through babel-js.

Still need to figure out media queries. Thoughts so far? I couldn't figure out a good way to reuse the existing code, since so much has changed. Since it is far more convenient in code, I am using the `:hover` syntax instead of state -> hover.


ianobermiller:
Also one thing I still need to fix here is that styles need to be applied in the order they are defined, so if `':hover'` is after `':active'`, it should take precedence. Right now they are hardcoded.

ianobermiller:
Ok, right now this is working as a drop-in replacement for Cesium. Yay! also the examples page works, with media queries.

colinmegill:
Nice!!

ianobermiller:
@alexlande, @colinmegill, what do you think? I've finished the work for media queries, which can include pseudo selectors. I've also added comprehensive tests, bringing the coverage of the core piece to 100%.

ianobermiller:
Also, this PR closes #80 and #82.

alexlande:
This is looking awesome :+1:, you are a hero, sir.

I'm still putting it through the paces, but am generally not seeing any serious issues. A few things:

- Can you clean up the code style a bit to get this passing lint? `gulp check` for eslint (should really just move that to an npm script but whatever).
- Can you replace the function declarations with function expressions?

Let's keep this branch open + get rewritten docs + guides in at the same time.

alexlande:
(I can jump in to on the code style stuff if you want too, just let me know)

modules/resolve-styles.js
```diff
@@ -0,0 +1,238 @@
...
+
+function _setStyleState(component, key, newState) {
+  var existing = component.state && component.state._radiumStyleState || {};
+  var state = { _radiumStyleState: clone(existing) };
+  state._radiumStyleState[key] = state._radiumStyleState[key] || {};
+  merge(state._radiumStyleState[key], newState);
+  component.setState(state);
+}
+
+// Merge style objects. Special casing for props starting with ';'; the values
```
alexlande: This should read `props starting with ':'`, right?

alexlande:
@ianobermiller: Any thoughts on how to handle named media queries with this rewrite? Don't want to lose that functionality if we can avoid it.

colinmegill:
@alexlande Happy to help with guides if you need it. We should probably also update the blog post.

colinmegill:
@ianobermiller great work - excited to use this

modules/resolve-styles.js
ianobermiller: Yes, should be "props starting with : or ("

ianobermiller:
Agreed on the docs and the site, and yes I can clean it up to pass lint.

As for named media queries, one thought is to leverage [computed property names](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Object_initializer#Computed_property_names), another way to say that Radium is Just JavaScript:

```javascript
var Breakpoints = {
  phone: '(max-width: 320px)',
  tablet: '(max-width: 1000px)',
  desktop: '(max-width: 1200px)',
};

var style = {
  width: '20%',
  [Breakpoints.phone]: {
    width: '100%'
  }
};

**Comment by alexlande** [comment]
Computed property names was my first though too, but I think it would be a bummer to take that functionality away unless you're using Babel.

In principle I agree with the "Radium is Just JavaScript" message. And maybe just making it less convenient to write style objects for this one use case if you're not using Babel isn't a huge deal.

**Comment by ianobermiller** [comment]
Is it worth increasing the API surface? We could do something like:

```javascript
Radium.defineMediaQuery('phone', '(max-width: 320px)');

var style = {
  width: '20%',
  '(phone)': {
    width: '100%'
  }
};
```

The parenthesis help distinguish it as a media query still.

examples/components/button.jsx
```diff
@@ -1,95 +1,87 @@
...
+    },
+    ':focus': {
+      backgroundColor: "#FF0000"
+    },
+    ':active': {
+      backgroundColor: "#990000"
+    }
+  },
+
+  mediaQueries: [
```
alexlande: Should be deleted.

alexlande:
That feels non-obvious to me, but I can't think of better syntax that wouldn't include a `mediaQueries` object, so maybe yeah?

alexlande:
Actually, now that I think about it, `(` as an indicator for media queries is sort of a problem. Media queries include more than just parentheticals: `@media tv and (min-width: 700px) and (orientation: landscape)`.

Maybe we make `@media` the prefix for a query, and do some other naming convention for custom ones?

Maybe `(--small)` as in the CSS custom media queries spec? http://dev.w3.org/csswg/mediaqueries/#custom-mq

ianobermiller:
@alexlande regarding reusing media queries, are computed properties good enough for now? Or do you think it is worth implementing the below? I don't love that it makes Radium more globally stateful.

```js
Radium.defineCustomMedia('small', 'max-width: 320px');

var style = {
  width: '20%',
  '@media (--small)': {
    width: '100%'
  }
};
```

alexlande:
Let's stick with computed properties for now and can think about custom queries for later.

(P.S. I'll aim to have PRs for docs + the website in today)

alexlande:
Nevermind, I see that you updated the docs. :fireworks:

Reviewing now and will leave comments here.

README.md
```diff
@@ -25,73 +25,85 @@ When we say expressive, we mean it: math, concatenation, regex, conditionals, fu
...
 - [API Docs](docs/api)
 
 ## Usage
 
-Start by writing a style object with a combination of default styles, browser states, media queries, and modifiers. Pass the object to `this.buildStyles()` and Radium will determine the correct group of style rules to apply to your component.
+Start by adding `Radium.wrap()` around the config you pass to `React.createClass`. Then, write a style object as you normally would with inline styles, and add in styles for interactive states and media queries. Pass the style object to your component via `style={...}` and let Radium do the rest!
 
 ```xml
-<Button kind='primary'>Radium Button</Button>
+<Button kind="primary">Radium Button</Button>
```
alexlande: Can we leave this as single quotes for consistency?

docs/api/README.md
```diff
@@ -4,226 +4,71 @@
...
-    }
-  ],
 
-  // Default media queries
-  mediaQueries: [
-    {
-      small: {
-        padding: 10
-      }
+    '@media (min-width: 1200px)': {
```
alexlande: Can we add an example of `:hover` *inside* a media query to illustrate how that works?

docs/api/README.md
```diff
@@ -4,226 +4,71 @@
...
+- Clean up any resources when the component unmounts
 
-## MatchMediaItem
+Usage is simple:
 
-The match media item mixin applies media queries set in `MatchMediaBase` to a component. To use, add it as a mixin to any component that is a descendent of the component with `MatchMediaBase` that should include media query styles.
+```javascript
+var MyComponent = React.createClass(Radium.wrap({
+  // write your component as normal
+}));
```
alexlande: This `pre` block needs to be closed.

alexlande:
Looks rad. Couple of minor issues, but otherwise I think this thing is ready to merge.

I'm thinking lets:

1. Merge this
2. I'll update the website with the new content
3. Release a new version

Then we can start getting the features in your other PRs in?

README.md
ianobermiller: Consistency with normal JS strings? FWIW, Facebook convention for JSX is to always use double quotes for string attributes. I'll change it back for now.

ianobermiller: Actually, we should pick a convention, because elsewhere in the docs we use double quotes in JSX (docs for `<Style>`)

docs/api/README.md
ianobermiller: Sure thing!

docs/api/README.md
ianobermiller: Good catch.

ianobermiller:
Sounds great! Let's do it!

README.md
alexlande: Oh jeez. Interesting re: Facebook convention. I'd prefer to stick with all single quotes everywhere, rather than using a different type depending on the context.

That said, this is a nit that can be addressed later :+1:

ianobermiller: Yeah, I think it stems from the HTML convention to use double quotes for attributes, so it always looks wierd to see JSX, which is like HTML, with single quotes.

alexlande:
The build failure is an eslint version issue (a rule changed names). Only present if you use `gulp check` instead of the `npm run` script. Going to merge this and remove gulp after.

ianobermiller:
Thanks, I was wondering how we solve that.

PR #108, opened 2015-04-22
Fix for 'document is undefined' error when running in node.

benoneal:


modules/mixins/match-media-base.js
```diff
@@ -54,6 +54,10 @@ var MatchMediaBase = {
   },
 
   handleMediaChange: debounce(function () {
+    if (typeof window !== 'undefined') {
```
ryan-roemer: Would we be better off with routing everything through `canUseDOM`? https://github.com/facebook/react/blob/master/src/vendor/core/ExecutionEnvironment.js

alexlande:
We should probably use `canUseDOM`, although we aren't doing that in the other place we do this check so maybe it's best to stick to this for now for consistency.

That said, shouldn't we not be hitting this code path on the server?

We don't initialize media queries on node: https://github.com/FormidableLabs/radium/blob/9290888912c031d5385baf09f2a851de623e6d86/modules/mixins/match-media-base.js#L25-L27

benoneal:
It's getting called from the second line of componentWillMount, which executes during React.renderToString().

alexlande:
Resolved by #112

alexlande:
Thanks for the patch though! Sorry for the late resolution. We're about to land a new API that does away with all of this behavior so it's been weird juggling the existing issues with the new behavior.

Issue #11, opened 2015-01-10
Vendor prefixes


alexlande:

mattborn:
I am so ready to start using this, but vendor prefixes were my third major concern after pseudo states and media queries. Not sure what the thinking is on this so far, but perhaps radium could work with [Autoprefixer](https://github.com/postcss/autoprefixer) somehow so it’s not a concern at all?

pstoica:
Maybe you could reuse JSS's css-vendor module? It checks on the client and caches the proper prefixes. Less bloat to download, and it's not expensive for browsers to do. https://github.com/jsstyles/css-vendor

ianobermiller:
Something like prefixfree might work as well: https://github.com/LeaVerou/prefixfree

alexlande:
Definitely agree that this should work transparently, ideally with Can I Use... support like Autoprefixer. Also ideally we wouldn't write this ourselves. Probably need to start by doing a survey of the solutions listed here (and any others we can think of) to see what might work best.

gpbl:
material-ui is also [porting](https://github.com/callemall/material-ui/tree/css-in-js) its CSS to JS. For what I see, they are using [a mixin](https://github.com/callemall/material-ui/blob/css-in-js/src/js/mixins/style-propable.js#L13) which adopted [a custom version](https://github.com/callemall/material-ui/blob/css-in-js/src/js/utils/modernizr.custom.js) of modernizr. It could be a source of inspiration. Maybe @hai-cea could drop some tip :-)

mattborn:
@pstoica @ianobermiller My gut tells me it’s worth it to avoid a client-side solution even if it’s not expensive in the browser. I don’t have a scientific reason, but some of the team members I work with prefer to declare exactly which browsers we support and have the vendor prefixes explicitly rolled up in the build.

Even though Autoprefixer is a CSS parser, I was hoping it would be easy to implement since it’s already 98.2% Javascript. That said, I am a product designer + don’t know that I’d be much help with the level of neck-bearded thinking that will be required to figure this out.

hai-cea:
Modernizr comes with a handy function that will prefix js styles only if the browser needs it:
http://modernizr.com/docs/#prefixed

We generated a custom Modernizer build by only selecting the tests we needed:
http://modernizr.com/download/

We also created some simple helper functions to prefix our style objects:
https://github.com/callemall/material-ui/blob/css-in-js/src/js/styles/auto-prefix.js

[...]

PR #109, opened 2015-04-22
Vendor prefixing

ianobermiller:
Add some simple prefixing capability, based on [css-vendor](https://github.com/jsstyles/css-vendor). Don't use css-vendor directly, though, to avoid going back and forth between dash-case and camelCase. Also, be smarter than css-vendor about valid values that the browser rewrites.

While this approach works quite well, it falls down a bit for values that require complex prefixing. [PrefixFree](https://github.com/LeaVerou/prefixfree/) is much more comprehensive. For example, it will prefix values inside of `transition`, rewrite `linear-gradient`, and more.

### DIFF
--- a/modules/__mocks__/prefix.js
+++ b/modules/__mocks__/prefix.js
@@ -1,0 +1,7 @@
+'use strict';
+
+var prefixMock = function(style, mode) {
+  return style;
+};
+
+module.exports = prefixMock;
--- a/modules/prefix.js
+++ b/modules/prefix.js
@@ -3,7 +3,8 @@
  * shenanigans.
  */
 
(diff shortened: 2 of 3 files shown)


alexlande:
I really want Radium to handle vendor prefixing, and there's obviously a huge demand for it given #11, but I'm not sure about implementing browser-only prefixing.

On the one hand, we already have several features that are browser-only at the moment, like media queries + pseudo-selectors. On the other, I desperately want to get both of those features working for server-side rendering.

I think where I'm landing on it is-- if we can come up with a plan to add isomorphic autoprefixing in a timely manner, we should do it. If not, I guess it can be one more thing on the list for when we fully support server-rendering.

PR #111, opened 2015-04-23
Add css keyframes animation helper

### LINKED ISSUE #52: Add a way to define keyframe css animations
It would be nice to be able to specify [keyframe animations](https://developer.mozilla.org/en-US/docs/Web/CSS/@keyframes) for an element in the component itself in javascript.

Thanks for this awesome library!

ianobermiller:
Fixes #52. Very simple to use:

```javascript
var pulseAnimation = Radium.animation({
  '0%': {width: '10%'},
  '50%': {width: '50%'},
  '100%': {width: '10%'}
});

var styles = {
  root: {
    animation: pulseAnimation + ' 3s ease 0s infinite'
  },
};
```

### DIFF
--- a/modules/index.js
+++ b/modules/index.js
@@ -2,4 +2,5 @@
 
 exports.Style = require('./components/style');
 exports.getState = require('./get-state');
+exports.keyframes = require('./keyframes');
 exports.wrap = require('./wrap');
--- a/modules/keyframes.js
+++ b/modules/keyframes.js
@@ -1,0 +1,57 @@
+'use strict';
+
+var prefix = require('./prefix');
+
+var kebabCase = require('lodash/string/kebabCase');
+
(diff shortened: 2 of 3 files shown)


ianobermiller:
Does anyone else think this should be called `Radium.keyframes` instead?

colinmegill:
Yes

ianobermiller:
Rename done! Fixed up all my crazy commits caused by using the Sync button in GitHub for Mac.

alexlande:
@ianobermiller: Any way we can get this in without depending on `vendor-prefix`? I'd like to mull that one over a bit more.

ianobermiller:
We could, but it wouldn't be terribly useful, as @keyframes still requires a prefix on Chrome and Safari.

PR #104, opened 2015-04-07
Make sure that handleMediaChange is called when component will be mounte...

brochington:
...d. Fixes #96

PR #112, opened 2015-04-24
fixed retrieving initial value being delayed, set default matchers for s...

taak77:
...erver side Fixes #104.


alexlande:
Awesome, thank you!

Issue #123, opened 2015-04-28
New API swaps abstraction for glue code.


benoneal:
Compared the the old API, the new API requires a LOT of handholding, with component render methods being filled with glue code which was neatly abstracted away in the old API, and the wrap implementation now makes abstraction layers between the app and your library impossible.  

In the old API, I had implemented a wrapping mixin which allowed me to do the following for all components: include the mixin, attach styles to the components' static method, and apply styles via my custom this.css([namespace]) method (which allowed me to chain things like vendor prefixing). Styles themselves declared what they looked like based on the state and props of the component with logical nesting. 

In the new API, I have to wrap each component in radium.wrap, attach styles as a variable, then glue state and props to the styles with a ton of completely inane "if blue, use blue" code, then wrap those in any processing methods I require, then apply them to elements via styles prop. Styles themselves can still (probably) declare what they look like based on state and props with nesting. 

Also, importantly for our app, we strongly prefer to create abstraction layers between libraries and the app (primarily through mixins), so that if we in future decide to switch libraries, or if a library introduces breaking changes to their API, we can simply modify the point of abstraction instead of having to manually propagate changes throughout hundreds of components. By requiring every component to be wrapped by radium, now my app has explicitly become a "radium app". I can't swap libraries without a complete sweep of all components. If you introduce breaking changes to how styles are applied, guess what? I have to manually change every single component. 

Both of these changes represent a severe regression for us. We are explicitly trying to abstract away trivial glue code and minimise the interface surface area with external libraries to promote rapid development which can be kept up-to-date with minimal effort.

benoneal:
Here is an example of our current implementation of radium in our production app, using an atomic 'input' component (with irrelevant code collapsed removed): 

Mixin:
```
var _ = require('lodash');
var prefix = require('react-prefixr');
var { StyleResolverMixin, BrowserStateMixin, MatchMediaItem } = require('radium');

module.exports = _.merge({
  css: function(selector, mods) {
    return prefix(this.buildStyles(this.constructor.style[selector], mods));
  },
  attachEvents: BrowserStateMixin.getBrowserStateEvents,
  Events: BrowserStateMixin // must add: { ...this.attachEvents() } to top level element
}, StyleResolverMixin, MatchMediaItem);
```

Input Component:
```
var React = require('react/addons');
var Style = require('../../../lib/mixins/style');

module.exports = React.createClass({
  displayName: "Input",

  mixins: [Style, Style.Events ...],

[...]

PR #133, opened 2015-05-07
Add Enhancer for using Radium with ES6 classes

### LINKED ISSUE #122: Using Radium with ES6 components
Hi, 

I wrote a class that when inherited allows creating component in an ES6 fashion. This class triggers different parts of Radium.wrap during its lifecycle, so it's quite dirty but here it is :

```javascript
import {Component} from 'react';
import {wrap} from 'radium';

class RadiumComponent extends Component {

    constructor(props, opts) {
        super(props, opts);
        if (this.state) {
            this.state._radiumStyleState = {};
        } else {
            this.state = { _radiumStyleState: {} };
        }
    }

    componentWillUnmount() {
        super.componentWillUnmount();

        if (this._radiumMouseUpListener) {
            this._radiumMouseUpListener.remove();
        }

        if (this._radiumMediaQueryListenersByQuery) {
            Object.keys(this._radiumMediaQueryListenersByQuery).forEach(
                function (query) {
                    this._radiumMediaQueryListenersByQuery[query].remove();
                },
                this
            );
        }
    }

    renderWithStyles() {
        console.error("renderWithStyles should be defined by your component");
        return null;
    }

    render(){
        var renderedElement = this.renderWithStyles();
        return wrap({render: () => renderedElement }).render();
    }
}

export default RadiumComponent ;
```

Your component just needs to implement a renderWithStyles method.
I'm not doing a PR as this is clearly a draft, and I don't know if you're interested in this in the short-term.
Including it in the project would allow for a cleaner implementation, since hacky calls to wrap would not be needed anymore.

Moreover, I don't know if classes and modules works well with JSX-loader.

ianobermiller:
I'm not really sure about the naming conventions, but this is a "higher order component". I've also played around with the idea of calling it a Decorator, since it can be used as-is as an ES7 decorator (https://github.com/wycats/javascript-decorators). We might also consider exposing `wrapUtils` (to satisfy cases like #123), although I'm not sure it is worth expanding the API surface.

Fixes #122.

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,6 +35,7 @@ When we say expressive, we mean it: math, concatenation, regex, conditionals, fu
 * Media queries
 * Automatic vendor prefixing
 * Keyframes animation helper
+* ES6 class and `createClass` support
 
 ## Docs
 
@@ -43,7 +44,7 @@ When we say expressive, we mean it: math, concatenation, regex, conditionals, fu
 
 ## Usage
 
-Start by adding `Radium.wrap()` around the config you pass to `React.createClass`. Then, write a style object as you normally would with inline styles, and add in styles for interactive states and media queries. Pass the style object to your component via `style={...}` and let Radium do the rest!
+Start by adding `Radium.Enhancer()` around your component, like `module.exports = Radium.Enhancer(Component)`, or `Component = Radium.Enhancer(Component)`. Alternatively, if using `createClass`, add `Radium.wrap()` around the config you pass to `React.createClass`. Then, write a style object as you normally would with inline styles, and add in styles for interactive states and media queries. Pass the style object to your component via `style={...}` and let Radium do the rest!
 
 ```as
 <Button kind="primary">Radium Button</Button>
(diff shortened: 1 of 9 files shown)


alexlande:
Finally played around with this and love it. Feels much nicer than the wrap method. Sorry about the merge conflict I introduced with the `keyframes` thing. :shipit:

ianobermiller:
How about naming? @colinmegill any comments here?

Some options:
- Enhancer
- Decorator
- Wrapper (tad confusing since we have `wrap`, but also maybe consistent?)
- HigherOrderComponent (far too wordy)

ianobermiller:
I'll add more tests for Enhancer once we have babel compilation support; it is hard to do stuff with classes without the class syntax :)

ianobermiller:
I'll merge this now, and we can change the name afterward if we so desire.

alexlande:
Posthumous comment: I like Enhancer. Lines up with a lot of the other higher order component discussions I've seen. `Decorator` implies that it is specifically an ES7 decorator, which would be confusing. `Wrapper` is too similar to `wrap`. `HigherOrderComponent` is indeed too long.

PR #140, opened 2015-05-12
Don't use React.Children.map if only one child

### LINKED ISSUE #139: Radium.wrap breaks React.Children.only
Using `Radium.wrap` breaks [elements requiring only one child](https://facebook.github.io/react/docs/top-level-api.html#react.children.only). The example I encountered this in is [react-bootstrap's `OverlayTrigger` component](https://github.com/react-bootstrap/react-bootstrap/blob/master/src/OverlayTrigger.js). Including this or similar elements anywhere in render will throw:

```
Uncaught Error: Invariant Violation: onlyChild must be passed a children with exactly one child.
```

ianobermiller:
Fixes #139.

Also makes it easier to write tests by using JSX.

### DIFF
--- a/jest/preprocessor.js
+++ b/jest/preprocessor.js
@@ -1,0 +1,6 @@
+var ReactTools = require('react-tools');
+module.exports = {
+  process: function(src) {
+    return ReactTools.transform(src);
+  }
+};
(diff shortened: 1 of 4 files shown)


alexlande:
Looks great!

PR #142, opened 2015-05-13
Bug: Make Children.only check handle string children

alexlande:
This is embarrassing. I pulled + ran the tests on https://github.com/FormidableLabs/radium/pull/140 but didn't run the examples. If I did, I would have noticed that it broke when the old child of a component is a string, not an element. This adds a check + a test for that case.

/cc @ianobermiller


alexlande:
Failed lint. Hanging my head in shame.

PR #144, opened 2015-05-13
Feature: Add babel compilation on prepublish

alexlande:
This revisits the compiling Radium with babel thing. Includes a set up to compile our modules to a `.gitignore`'d `lib` directory.

We can either do this on `prepublish` or in a `publish.sh`. I wasn't really sure what the best option was. The weird thing about `prepublish` is that it runs on `npm install`, not just before the package is published. See https://github.com/npm/npm/issues/3059 for a very long discussion about that.

If we do something like a `publish.sh` script that includes `npm run lib` and `npm publish`, we wouldn't slow down `npm install` for everyone.

If anyone has opinions on that...

/cc @ianobermiller @kenwheeler

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,7 +2,7 @@
   "name": "radium",
   "version": "0.11.1",
   "description": "A set of tools to manage inline styles on React elements",
-  "main": "modules/index.js",
+  "main": "lib/index.js",
   "repository": {
     "type": "git",
     "url": "https://github.com/formidablelabs/radium.git"
@@ -14,6 +14,7 @@
   },
   "scripts": {
     "dist": "webpack && webpack --config=webpack.config.minified.js",
+    "lib": "rm -rf ./lib && babel ./modules -d lib && rm -rf ./lib/__tests__ ./lib/__mocks__",
     "examples": "webpack-dev-server --config examples/webpack.config.js --no-info --content-base examples/",
     "test": "jest",
     "lint": "eslint modules"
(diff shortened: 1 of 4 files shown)


ianobermiller:
I strongly prefer `npm run dist` or `npm run lib` over prepublish. The onus should never be on installers to compile.

kenwheeler:
I put it in as a safeguard to make sure the most recent build has taken place prior to a publish, but I'm not so hot on it running on install like that

alexlande:
Cool, I was leaning away from `prepublish`. Going to merge this in as is for now, then we can revisit whether we need a script to do `lib` + `publish` (+ anything else?) a bit later.

package.json
```diff
@@ -14,6 +14,7 @@
   },
   "scripts": {
     "dist": "webpack && webpack --config=webpack.config.minified.js",
+    "lib": "rm -rf ./lib && babel ./modules -d lib && rm -rf ./lib/__tests__ ./lib/__mocks__",
```
ianobermiller: use rimraf instead of rm -rf so it works on windows (not sure if there are other blockers, but no need to add new ones)

ryan-roemer: Also, as `GLOBAL` comment, remove all `./` prefixes as Unix/Mac doesn't need them and they break Windows.

alexlande:
Will do re: `rimraf` + `./`

One other issue that I hit with this last night is that npm installing a particular git hash of Radium no longer works (unless you install and then run `npm run lib` in your project that you're installing it into).

I don't know how important that use case is, but it's something I do semi-regularly to test/use Radium updates before we release them.

doctyper:
@alexlande +1. I haven't seen anything that discourages `precompile` scripts. I think it's the recommended approach when using Babel: http://mammal.io/articles/using-es6-today/

ianobermiller:
Sounds fine to me, the hash checkout is compelling.

PR #146, opened 2015-05-13
Replaceable matchMedia() function for server side rendering

azazdeaz:
Hi! We are using this fork both for server side rendering and generating mobile device preview on client side.

It replaces the ```window.matchMedia()``` calls with ```Radium.matchMedia()``` which is working like a strategy so you can replace the currently used matchMedia function like ```Radium.matchMedia.replace(iframe.contentWindow.matchMedia)```

It provides [css-mediaquery](https://github.com/ericf/css-mediaquery) based substitution for ```window.matchMedia()``` (```MatchMediaMock```) so it can be used with only js on the server side.

ex:
```javascript
var matchMediaMock = require('radium/modules/match-media-mock');
Radium.matchMedia.replace(matchMediaMock);

app.get('/app/:width/:height', function(req, res) {
  
  matchMediaMock.setConfig({
    type: 'screen',
    width: req.params.width,
    height: req.params.height,
  });
  
  var html = React.renderToString(<RadiumApp/>);
  
  res.end(html);
});
```
As you see, it's not a full solution for the server side rendering because you have to be able to send some information about the client device in the request. I don't know whether this use case support is among your goals but if so, and you like this approach i would be happy to prepare this to be mergeable.

### DIFF
--- a/modules/resolve-styles.js
+++ b/modules/resolve-styles.js
@@ -5,6 +5,7 @@
 var MouseUpListener = require('./mouse-up-listener');
 var getState = require('./get-state');
 var Prefixer = require('./prefixer');
+var Config = require('./config');
 
 var ExecutionEnvironment = require('exenv');
 var React = require('react');
@@ -75,11 +76,7 @@ var _onMediaQueryChange = function (component, query, mediaQueryList) {
 };
 
 var _resolveMediaQueryStyles = function (component, style) {
-  if (
-    !ExecutionEnvironment.canUseDOM ||
-    !window ||
-    !window.matchMedia
-  ) {
+  if (!Config.canMatchMedia()) {
     return style;
   }
 
@@ -92,7 +89,7 @@ var _resolveMediaQueryStyles = function (component, style) {
     // Create a global MediaQueryList if one doesn't already exist
     var mql = mediaQueryListByQueryString[query];
     if (!mql) {
-      mediaQueryListByQueryString[query] = mql = window.matchMedia(query);
+      mediaQueryListByQueryString[query] = mql = Config.matchMedia(query);
     }
 
     // Keep track of which keys already have listeners
--- a/modules/__tests__/resolve-styles-test.js
+++ b/modules/__tests__/resolve-styles-test.js
@@ -5,11 +5,13 @@
 
 jest.dontMock('../get-state.js');
 jest.dontMock('../resolve-styles.js');
+jest.dontMock('../config.js');
 
 var React = require('react');
 var MouseUpListener = require('../mouse-up-listener.js');
 var objectAssign = require('object-assign');
 var resolveStyles = require('../resolve-styles.js');
+var Config = require('../config.js');
 
 var genComponent = function () {
   return {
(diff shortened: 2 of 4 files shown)


alexlande:
Thanks @azazdeaz, this looks cool. In general, I don't think that we want to include something like `match-media-mock` directly, but allowing an interface like `Radium.matchMedia.replace()` for this use case makes sense to me.

Long term, I still think we need to do CSS generation for media queries + states as in https://github.com/FormidableLabs/radium/issues/53 for proper isomorphic rendering, but this would help in the meantime.

/cc @ianobermiller, thoughts?

ianobermiller:
Yes, I like this. I'll add a bunch of comments in code, but I think a better and more scalable interface might be to expose a `Radium.Config` object with a bunch of methods. In this case, it would be `Radium.Config.setMatchMedia`. We've also talked about adding support for named media queries, which would fit nicely here.

modules/keyframes.js
```diff
@@ -3,13 +3,14 @@
 var prefix = require('./prefix');
 
 var kebabCase = require('lodash/string/kebabCase');
+var isNode = require('detect-node');
```
ianobermiller: Instead of `detect-node`, we should use a copy of React's execution environment (since we shouldn't be deep linking into React): https://github.com/facebook/react/blob/2aeb8a2a6beb00617a4217f7f8284924fa2ad819/src/vendor/core/ExecutionEnvironment.js

modules/match-media-mock.js
```diff
@@ -0,0 +1,86 @@
...
+var pull = require('lodash/array/pull');
+var mapValues = require('lodash/object/mapValues');
+var forOwn = require('lodash/object/forOwn');
+var mediaQuery = require('css-mediaquery');
+var isNode = require('detect-node');
+
+var config = {};
+var createdMqls = {};
+
+function matchMediaMock(query) {
```
ianobermiller: as @alexlande said, this isn't bad, but I'd like to figure out our isomorphic solution before we add this specific of mock code and take a dependency on css-mediaquery

modules/resolve-styles.js
```diff
@@ -51,7 +52,7 @@ var _mouseUp = function (component) {
 
 var _onMediaQueryChange = function (component, query, mediaQueryList) {
   var state = {};
-  state[name] = mediaQueryList.matches;
+  state[query] = mediaQueryList.matches;
```
ianobermiller: good catch, we should fix this anyway. I wonder why lint didnt catch..

azazdeaz:
Hi! I rebased and made the changes regards your comments. Now the example looks like this: 
```javascript
var matchMediaMock = require('match-media-mock').create();
Radium.Config.setMatchMedia(matchMediaMock);

app.get('/app/:width/:height', function(req, res) {

  matchMediaMock.setConfig({
    type: 'screen',
    width: req.params.width,
    height: req.params.height,
  });

  var html = React.renderToString(<RadiumApp/>);

  res.end(html);
});
```

And here you can check the match-media-mock module: https://github.com/azazdeaz/match-media-mock

ianobermiller:
This looks great, thank you!

bunkat:
This pull request broke IE11.  The `matchMedia` export overrides the built-in `matchMedia` function which causes IE to throw an `Invalid calling object` exception and will not load the page.

Issue #141, opened 2015-05-12
Release 0.12


alexlande:
We've got several new features and some bug fixes in. I think it's about time for 0.12.

@ianobermiller: Anything else you think we should get in beforehand? If not will release today.

alexlande:
Actually-- it would be really nice to get the Enhancer in before-hand. Will take that for a test drive today, but it's looking good to me.

mrtnbroder:
I just installed from master – does radium not support isomorphic apps? I get the following error: 

```
/Users/mbroder/Sites/prototype/node_modules/radium/modules/prefix.js:19
var domStyle = document.createElement('p').style;
               ^
ReferenceError: document is not defined
    at Object.<anonymous> (/Users/mbroder/Sites/prototype/node_modules/radium/modules/prefix.js:19:16)
    at Module._compile (module.js:428:26)
    at Object.Module._extensions..js (module.js:446:10)
    at Module.load (module.js:353:32)
    at Function.Module._load (module.js:308:12)
    at Module.require (module.js:363:17)
    at require (module.js:382:17)
    at Object.<anonymous> (/Users/mbroder/Sites/prototype/node_modules/radium/modules/keyframes.js:3:14)
    at Module._compile (module.js:428:26)
    at Object.Module._extensions..js (module.js:446:10)
```

alexlande:
We do. There are a couple of other issues holding up releasing the current stuff as well (see https://github.com/FormidableLabs/radium/pull/142 also).

Will plan to release by the end of the week.

mrtnbroder:
Oh, alright. Nvm me then. Keep up the good work!

awkaiser:
Cool, I'm looking forward to next NPM release with `Radium.Enhancer`, _and_ an example of what needs to be done for isomorphic apps (#143).

I had Radium working just fine in an isomorphic app roughly one month ago, but that code broke at some point during the transition from mixins to wrapped components. I don't mind, because `Enhancer` is a welcomed change and I'm not expecting a stable public API until 1.0 (updating my code to use `Enhancer` was easy), but this usage of `document` methods outside the browser is frustrating.

ianobermiller:
We should be more diligent about touching browser-only methods. The test
tool, jest, doesn't help since it includes jsdom by default.

ryan-roemer:
I'd suggest just adding some straight Node + mocha tests...

alexlande:
[...]

PR #149, opened 2015-05-14
Use exenv to bail on browser-specifics

ianobermiller:
While rendering on the server, we cannot access browser specific features. Use `exenv` to check if the browser environment is available.

Required for #141.


alexlande:
Nice :+1:

PR #154, opened 2015-05-14
Check for an available DOM before accessing DOM methods.

doctyper:
This PR does some sanity checking to make sure a DOM exists before accessing DOM-specific methods, which allows Radium to run in isomorphic environments.


ianobermiller:
Thanks, I updated #149 with some ideas from here!

PR #155, opened 2015-05-15
Release v0.12.0

### LINKED ISSUE #141: Release 0.12

### LINKED ISSUE #145: Where is Radium.getState?
I do

````
````

but there is no Radium.getState

### PR DESCRIPTION
**Pull request by ianobermiller**
closes #141
closes #145

### CODE REVIEW

File: modules/resolve-styles.js [hunk 1]
```diff
@@ -5,7 +5,7 @@ var getState = require('./get-state');
 var prefix = require('./prefix');
 
 var ExecutionEnvironment = require('exenv');
-var React = require('react/addons');
```
ianobermiller: My bad! This was causing us to suck react into the dist build!

awkaiser:
@ianobermiller the **package.json** should have `"main": "modules/index.js"` instead of `"main": "lib/index.js"`

awkaiser:
Ah, I see, the Babel transpiled modules end up in `lib` via `npm run lib`. Nevermind. :)

Giving this a whirl...

awkaiser:
@ianobermiller & @alexlande: [As supported by](https://github.com/FormidableLabs/radium/pull/144#issuecomment-102191039) @doctyper, why not use [NPM prepublish + **.npmignore**](http://mammal.io/articles/using-es6-today/)? IMHO, NPM packages should be immediately usable via the entry script given set as `main` in **package.json**.

doctyper:
Agreed, it would be great if you could get that and the [`rimraf`](https://github.com/FormidableLabs/radium/pull/144#discussion_r30285056) change in 0.12.

ianobermiller:
@awkaiser I thought that was what we were doing, I guess not! Yes, I think prepublish is the way to go.

PR #158, opened 2015-05-17
Remove lodash to make the distributed build smaller

ianobermiller:
- Use object-assign instead of lodash merge
- Replace generic kebabCase with a more compact and specific version
- Reduces gzipped size of Radium distribution from 10,366 bytes to 4,093 bytes

```bash
~/dev/radium git master
➜ gzip < dist/radium.min.js | wc -c
   10366

~/dev/radium git reduce-size
➜ gzip < dist/radium.min.js | wc -c
    4093
```


alexlande:
Nice, given how much simpler the style object resolution is now, we definitely don't need `_.merge`.

:shipit:

PR #165, opened 2015-05-18
add fixes to Enhancer

azazdeaz:
Hi! I started to play with the new ```@Radium.Enhancer``` and i ran into a few issue: it isn't passing the props to the original element,  it calls the original constructor twice, and it hides the original ```displayName```.

I added the tests and the fixes.

I learnt the practices from [here](https://gist.github.com/sebmarkbage/ef0bf1f338a7182b6775) and [here](https://github.com/gaearon/react-dnd/blob/19f67400feaa168d97a667720c746eca1afab494/src/DragDropContext.js)

### DIFF
--- a/modules/__tests__/enhancer-test.js
+++ b/modules/__tests__/enhancer-test.js
@@ -38,6 +38,19 @@ describe('Enhancer', () => {
     );
   });
 
+  it('receives the given props', () => {
+    class Composed extends Component {
+      constructor (props) {
+        super(props);
+      }
+    }
+    var Enhanced = Enhancer(Composed);
+
+    var instance = new Enhanced({foo: 'bar'});
+
+    expect(instance.props).toEqual({foo: 'bar'});
+  });
+
   it('calls existing render function, then resolveStyles', () => {
     var renderMock = jest.genMockFunction();
     class Composed extends Component {
@@ -53,6 +66,30 @@ describe('Enhancer', () => {
 
     expect(renderMock).toBeCalled();
     expect(resolveStyles).toBeCalled();
+  });
+
+  it('calls existing constructor only once', () => {
+    var constructorMock = jest.genMockFunction();
+    class Composed extends Component {
+      constructor () {
+        super();
+        constructorMock();
+      }
+    }
+    var Enhanced = Enhancer(Composed);
+
+    new Enhanced(); // eslint-disable-line no-new
+
+    expect(constructorMock.mock.calls.length).toBe(1);
+  });
+
+  it('refers to the existing displayName', () => {
+    class Composed extends Component {}
+    Composed.displayName = 'Composed';
+
+    var Enhanced = Enhancer(Composed);
+
+    expect(Enhanced.displayName).toContain(Composed.displayName);
   });
 
   it('calls existing componentWillUnmount function', () => {
--- a/modules/enhancer.js
+++ b/modules/enhancer.js
@@ -1,47 +1,3 @@
-// 'use strict';
-//
-// var resolveStyles = require('./resolve-styles.js');
-// var wrapUtils = require('./wrap-utils.js');
-//
-// var enhanceWithRadium = function (ComposedComponent) {
-//   var RadiumEnhancer = function () {
-//     ComposedComponent.prototype.constructor.call(this);
-//
-//     if (!this.state) {
-//       this.state = {};
-//     }
-//
-//     var radiumInitialState = wrapUtils.getInitialState();
-//     Object.keys(radiumInitialState).forEach(function (key) {
-//       this.state[key] = radiumInitialState[key];
-//     }, this);
-//   };
-//
-//   RadiumEnhancer.prototype = new ComposedComponent();
-//   RadiumEnhancer.prototype.constructor = RadiumEnhancer;
-//
-//   RadiumEnhancer.prototype.render = function () {
-//     var renderedElement = ComposedComponent.prototype.render.call(this);
-//     return resolveStyles(this, renderedElement);
-//   };
-//
-//   RadiumEnhancer.prototype.componentWillUnmount = function () {
-//     if (ComposedComponent.prototype.componentWillUnmount) {
-//       ComposedComponent.prototype.componentWillUnmount.call(this);
-//     }
-//
-//     wrapUtils.componentWillUnmount(this);
-//   };
-//
-//   RadiumEnhancer.defaultProps = ComposedComponent.defaultProps;
-//   RadiumEnhancer.propTypes = ComposedComponent.propTypes;
-//   RadiumEnhancer.contextTypes = ComposedComponent.contextTypes;
-//
-//   return RadiumEnhancer;
-// };
-//
-// module.exports = enhanceWithRadium;
-
 'use strict';
 
 var resolveStyles = require('./resolve-styles.js');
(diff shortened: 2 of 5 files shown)


modules/enhancer.js
```diff
@@ -2,40 +2,39 @@
...
 var enhanceWithRadium = function (ComposedComponent) {
-  var RadiumEnhancer = function () {
-    ComposedComponent.prototype.constructor.call(this);
 
-    if (!this.state) {
-      this.state = {};
-    }
+  class RadiumEnhancer extends ComposedComponent {
+
+    constructor (props) {
```
ianobermiller: I don't think this is how classes work with React; afaik, there is nothing passed to the constructor.

modules/enhancer.js
```diff
@@ -2,40 +2,39 @@
...
 
 var enhanceWithRadium = function (ComposedComponent) {
-  var RadiumEnhancer = function () {
-    ComposedComponent.prototype.constructor.call(this);
 
-    if (!this.state) {
-      this.state = {};
-    }
+  class RadiumEnhancer extends ComposedComponent {
+
```
ianobermiller: Remove extra blank line

ianobermiller:
Cool stuff, please rebase onto master.

modules/__tests__/enhancer-test.js
```diff
@@ -38,6 +38,19 @@ describe('Enhancer', () => {
...
 
+  it('receives the given props', () => {
+    class Composed extends Component {
+      constructor (props) {
+        super(props);
+      }
+    }
+    var Enhanced = Enhancer(Composed);
+
+    var instance = new Enhanced({foo: 'bar'});
```
ianobermiller: tests should not use new; they should use JSX syntax

modules/enhancer.js
```diff
@@ -2,40 +2,39 @@
...
+  }
 
-    wrapUtils.componentWillUnmount(this);
-  };
+  const displayName =
+    ComposedComponent.displayName ||
+    ComposedComponent.name ||
+    'Component';
 
-  RadiumEnhancer.defaultProps = ComposedComponent.defaultProps;
```
ianobermiller: These should all be kept

modules/enhancer.js
```diff
@@ -2,40 +2,39 @@
...
-  RadiumEnhancer.propTypes = ComposedComponent.propTypes;
-  RadiumEnhancer.contextTypes = ComposedComponent.contextTypes;
+  RadiumEnhancer.displayName = `Radium(${displayName})`;
```
ianobermiller: move this to static on the class as react-dnd. In fact, all of these can be moved to static class properties.

modules/enhancer.js
ianobermiller: Oh yeah I am totally wrong here. The constructor should take, and pass along, props and context.

modules/enhancer.js
ianobermiller: In fact, we should be copying all static properties from the composted component.

modules/__tests__/enhancer-test.js
azazdeaz: Do you mean something like this
```var element = <Enhanced/>;
React.renderToString(element);```
instead of this
```var instance = new Enhanced();```?

File: modules/enhancer.js (hunk 5 above)
Comment (azazdeaz) [inline modules/enhancer.js]: Afaik, the static props should be inherited by default. [ie](http://babeljs.io/repl/#?experimental=true&evaluate=true&loose=false&spec=false&code=class%20A%20%7B%0A%20%20static%20X%20%3D%201%0A%7D%0A%0Aclass%20B%20extends%20A%20%7B%7D%0A%0Aconsole.log(B.X)%2F%2F1)

**Comment by azazdeaz** [comment]
Sorry, there is an error thrown after the tests passed, on the coverage html generating. I found some issue about it but i couldn't figure out what should i do with it jet. (I'm not familiar with Jest or Istanbul)

File: modules/__tests__/enhancer-test.js (hunk 3 above)
Comment (ianobermiller) [inline modules/__tests__/enhancer-test.js]: No, like this:

```as
var element = <Enhanced foo="bar" />;
expect(element.props).toEqual({foo: 'bar'});
```

ianobermiller:
You may be running into the jest issue where it needs to be run under an older version of Node; I use `nvm` to grab v0.10.x.

modules/enhancer.js
```diff
@@ -2,40 +2,37 @@
...
-    Object.keys(radiumInitialState).forEach(function (key) {
-      this.state[key] = radiumInitialState[key];
-    }, this);
-  };
+  class RadiumEnhancer extends ComposedComponent {
+    static displayName = `Radium(${displayName})`;
 
-  RadiumEnhancer.prototype = new ComposedComponent();
-  RadiumEnhancer.prototype.constructor = RadiumEnhancer;
+    constructor (props) {
```
ianobermiller: also pass context, or just change super to `super(...arguments)` (assuming rest params are supported by babel)

modules/enhancer.js
```diff
@@ -2,40 +2,37 @@
...
     }
 
-    wrapUtils.componentWillUnmount(this);
-  };
+    componentWillUnmount () {
+      if (super.componentWillUnmount) {
+        super.componentWillUnmount();
+      }
 
-  RadiumEnhancer.defaultProps = ComposedComponent.defaultProps;
```
ianobermiller: We still need to transfer all these static props, and should probably transfer any other props that someone has set (if they set custom statics)

azazdeaz:
>You may be running into the jest issue where it needs to be run under an older version of Node; I use nvm to grab v0.10.x.

No, the error is coming after the tests are passed from the coverage generation. Regards [this](https://github.com/babel/babel-jest/issues/11) and [this](https://github.com/facebook/jest/issues/304) the coverage generator is falling because the babel generated source is not the same as the original and loks like there is no workaround yet. So i think the best thing we can do is remove the coverage task from the basic npm test and add new srcript to npm (```npm run test-coverage````) whitch is runs the coverage test as before.

>We still need to transfer all these static props, and should probably transfer any other props that someone has set (if they set custom statics)

Sorry. i dont know where is my notes gone on this. Afaik they should be inherited by default. [for example](http://babeljs.io/repl/#?experimental=true&evaluate=true&loose=false&spec=false&code=class%20A%20%7B%0A%20%20static%20X%20%3D%201%0A%7D%0A%0Aclass%20B%20extends%20A%20%7B%7D%0A%0Aconsole.log(B.X)%2F%2F1)

>
```
var element = <Enhanced foo="bar" />;
expect(element.props).toEqual({foo: 'bar'});
```

It's tricky because ReactElements only wraps the Component class and don't instantiate it until it gets rendered.
[React (Virtual) DOM Terminology](https://facebook.github.io/react/docs/glossary.html)
[playground](http://codepen.io/azazdeaz/pen/doXQrp?editors=001)
So in this case the element.props would be equal to {foo: 'bar'} but it has nothing to do with the tested Component. But i think so that in some test cases it would be more convenient to test with real ReactElements.

ianobermiller:
Good to know, thanks for the info. I hadn't read the "React (Virtual) DOM Terminology" page since ES6 classes were introduced.

For coverage, let's have `test` not do coverage, and have a `test-coverage` as you mentioned, but run babel first and do the test on the `lib` directory. I'll set it up after this merges.

modules/enhancer.js
ianobermiller: Can you please add a test like the displayName one that propTypes an defaultProps are transferred correctly?

ianobermiller:
Other than those comments, just rebase and we'll get it merged. Thanks very much for your contribution!

azazdeaz:
It's fast forward. Thank for you for the handy and evolving lib!

PR #179, opened 2015-05-23
use styles[kind]

matjaz:


ianobermiller:
Not sure why I never thought of this before! Great idea!

Issue #127, opened 2015-04-30
Internet Explorer 9 support


cateland:
Using Radium with Internet Explorer generate the subsequent Error :

```SCRIPT438: L’objet ne gère pas la propriété ou la méthode « matchMedia » ```

Meaning that the window.matchMedia is not available.

I think it would be a great instead of crashing to notice the developper via a console warning in developpement mode that one of his component who use media query won't work fully as intended.
And give us an opportunitty to gracefully degrade such component.

ryan-roemer:
Generally speaking, with React my sense is that it's left up to the library consumer to polyfill down to the browsers you expect to support (e.g., `es5-shim`) -- https://facebook.github.io/react/docs/working-with-the-browser.html#browser-support-and-polyfills

Perhaps we just need to add similar documentation / warnings in the README / site pointing to http://caniuse.com/#feat=matchmedia and suggesting something like https://afarkas.github.io/webshim/demos/#matchMedia ?

cateland:
I tried to use https://github.com/paulirish/matchMedia.js polyfill to stumble on the next error (thx for pointing to that polyfill)

```SCRIPT438: Object doesn't support property or method 'addListener' ```

ryan-roemer:
`addListener` is going to need `es5-shim` polyfills IIRC

ryan-roemer:
Aha! https://github.com/paulirish/matchMedia.js/blob/master/matchMedia.addListener.js

alexlande:
Huh, I didn't realize that `addListener` was a separate polyfill. Sorry about that!

I think this is both a documentation + code issue. Radium definitely shouldn't blow up if a browser doesn't have `matchMedia` and you don't have a polyfill. It's valid to want to not support media queries on old IE but use them elsewhere, for example.

We need to:

- [ ] Update docs to mention this
- [ ] Handle cases `matchMedia` (and maybe `addListener` separately) aren't available.

MattKunze:
Keyframe animation doesn't work in IE9 either. I don't think there is a workaround or polyfill, but it would be nice to gracefully do nothing in this case rather than throwing exceptions.

I'm getting 'SCRIPT5022: DOM Exception: SYNTAX_ERR (12)' when it calls `animationStyleSheet.sheet.insertRule(rule, animationStyleSheet.sheet.cssRules.length)` in the `keyframes` function.

MattKunze:
[...]

PR #181, opened 2015-05-24
Fix IE9 when using keyframes

ianobermiller:
- Bail out on keyframes if animation not supported at all
- Refactor `prefix` into `Prefixer`
- References #127


alexlande:
:shipit:

PR #182, opened 2015-05-25
Fix react externals in webpack config

### LINKED ISSUE #178: Require 'react'
The current webpack ends up with `require("React")` which throws warnings on my system when I include it.  Should it be `require("react")`?

https://github.com/FormidableLabs/radium/blob/master/webpack.config.js#L8

ianobermiller:
Fixes #178

Copied from
https://github.com/rackt/react-router/blob/master/webpack.config.js


alexlande:
:shipit:

PR #186, opened 2015-05-30
Fix postinstall on windows

ianobermiller:
Creating a PR to test the appveyor build

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -6,6 +6,9 @@ environment:
 install:
   # Get the latest stable version of Node.js or io.js
   - ps: Install-Product node $env:nodejs_version
+  # Install and use local, modern NPM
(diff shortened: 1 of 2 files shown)


ryan-roemer:
@ianobermiller -- If appveyor doesn't like this another option is to just use `npm` programmatically in a Node script. Here's some totally unrelated code that looks for extra `devDependencies` that for Heroku only _should_ be installed and then uses NPM programmatically to install them... https://github.com/FormidableLabs/full-stack-testing/blob/master/app/heroku/install.js

(Just in case it helps ;) )

ianobermiller:
Probably running into this: https://github.com/joyent/node/issues/2318

ianobermiller:
Yes!

PR #190, opened 2015-06-06
Add note about IE9 and media queries

### LINKED ISSUE #127: Internet Explorer 9 support

ianobermiller:
- Fixes #127
- Also delete mis-spelled directory


ianobermiller:
Looks like a false negative on travis. Tested on IE9 using the polyfill and this bin http://jsbin.com/kotuhu/3/edit on http://cloudinternetexplorer.com/, and it works when you resize.

PR #191, opened 2015-06-06
Support fallback CSS values

### LINKED ISSUE #183: Support fallback values
See discussion in https://github.com/FormidableLabs/radium/issues/164#issuecomment-106547441.

Plan is to add this to the prefixer; user can specify an array of values, and each one in order will be tested and used if it works:

```as
{
  color: ['rgba(255, 255, 255, 50)', '#eee']
}
```

ianobermiller:
Fixes #183

### DIFF
--- a/docs/guides/README.md
+++ b/docs/guides/README.md
@@ -256,7 +256,7 @@ var styles = {
 
 You can query Radium's state using `Radium.getState`. This allows you to style or render one element based on the state of another, e.g. showing a message when a button is hovered.
 
-```js
+```as
 // Inside render
 return (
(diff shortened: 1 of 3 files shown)


alexlande:
Awesome. Still think this sort of needs to be in React core for server-side support, but it's great that we have something.

alexlande:
:+1:

sassanh:
Does it still work for you? It doesn't work here.

DanielSinclair:
Having issues with this — and am not able to find documentation. Has this feature been removed? Why?

PR #192, opened 2015-06-06
Style component should take an object, not an array

### LINKED ISSUE #153: Style component should take an object, not an array
Right now, the Style component takes an array of rules, which is technically correct but doesn't line up with the way we handle any of our other styles objects, which are all objects.

Should update `Style` so that it doesn't require such a verbose syntax.

ianobermiller:
Fixes #153


alexlande:
:shipit:

PR #193, opened 2015-06-06
Remove Radium.wrap

### LINKED ISSUE #188: Remove Radium.wrap
I propose we remove `Radium.wrap` for v0.13. You can use `Radium.Enhancer` on the return of `React.createClass` anyway. The only advantage of `Radium.wrap` is that it should be slightly more efficient than a Higher-Order Component. I've not measured any of it, and I suspect there are lower hanging performance fruit anyway.

Once wrap has been removed, I also propose we simplify the API by making the default export the higher-order function, so you can just write:

```as
// Decorator
@Radium
class App extends Component { ... }

// ES6 + wrapper
class App extends Component { ... }
var App = Radium(App);

// createClass + wrapper
var App = React.createClass({ ... });
var App = Radium(App);
```

ianobermiller:
Remove `Radium.wrap`, and change `Radium.Enhancer` to just `Radium`.
Update all the docs.
Closes #188


alexlande:
I've got to say, Radium as a decorator is absurdly beautiful. Great idea to promote it as the recommended way to use the lib. Everything here looks great :+1:

PR #194, opened 2015-06-07
Leave primitive children alone

### LINKED ISSUE #189: Browser will not render svg element when wrapped by Radium
For some reason, when I use Radium on the below React class, the browser does not render it (prob because it spits out invalid markup) then React gives me "Uncaught Error: Invariant Violation" because the dom and virtual dom don't match.

```
define([
  'React',
  'Radium',
  'underscore'
], function (React, Radium, _){

  var BbAppTreeNode = React.createClass({
    render: function(){
      return(
        <g className='node'>
          <circle r='4.5'></circle>
          <text x="-10" dy=".35em" textAnchor="end">{this.props.data.name}</text>
        </g>
      )
    }
  })

  return Radium.Enhancer(BbAppTreeNode)

})
```

I can get the class to render if:
1. I remove {this.props.data.name} from the text element. Leaving it empty.
2. I don't wrap it in Radium.Enhancer

I'm using Radium v0.12.0 and React v0.13.3

ianobermiller:
Fixes #189

### DIFF
--- a/modules/__tests__/resolve-styles-test.js
+++ b/modules/__tests__/resolve-styles-test.js
@@ -92,6 +92,22 @@ describe('resolveStyles', function () {
       expect(children[0].props.style).toBe(style);
     });
 
+    it('doesn\'t wrap string children in spans', function () {
+      var component = genComponent();
+      var renderedElement = <div>Hello</div>;
+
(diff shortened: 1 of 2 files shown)


alexlande:
:+1:

PR #202, opened 2015-06-09
fix resolving ReactComponentElement children

azazdeaz:
Currently Radium resolves all the childrens of the enhanced element which can lead to some unexpected behavior and affect the performance badly. 
ex:
```javascript
<MyRadiumComponent>
  <ThirdPartyComponent/>
  <div/>
</MyRadiumComponent>
```
The ```ThirdPartyComponent``` and it's children will be resolved, not just the ```div``` 

```javascript
<MyRadiumComponent>
  <MyRadiumComponent>
    <MyRadiumComponent/>
  </MyRadiumComponent>
</MyRadiumComponent>
```
The inner ```MyRadiumComponent```-s will be added to the ```resolveStyles()``` multiple times.

I added a failing test for this and a possible fix which is only resolve ```ReactDOMElement```-s when it recurse over the children in the ```resolveStyles()``` and skips the ``ReactComponentElement```-s.

### DIFF
--- a/modules/__tests__/resolve-styles-test.js
+++ b/modules/__tests__/resolve-styles-test.js
@@ -761,4 +761,18 @@ describe('resolveStyles', function () {
     });
   });
 
+  describe('ReactComponentElement children', function () {
+    it('doesn\'t change ReactComponentElement children', function () {
+      var component = genComponent();
+      class CustomComponent extends React.Component {}
+      var style = {':hover': {}};
+      var renderedElement = <div>
+        <CustomComponent style={style}/>
+      </div>;
+
+      var result = resolveStyles(component, renderedElement);
+      var children = getChildrenArray(result.props.children);
+      expect(children[0].props.style).toBe(style);
+    });
+  });
 });
--- a/modules/__tests__/resolve-styles-test.js
+++ b/modules/__tests__/resolve-styles-test.js
@@ -762,17 +762,48 @@ describe('resolveStyles', function () {
   });
 
   describe('ReactComponentElement children', function () {
-    it('doesn\'t change ReactComponentElement children', function () {
+    it('doesn\'t resolve ReactComponentElement children', function () {
       var component = genComponent();
       class CustomComponent extends React.Component {}
       var style = {':hover': {}};
-      var renderedElement = <div>
+      var renderedElement = (<div>
         <CustomComponent style={style}/>
-      </div>;
+      </div>);
 
       var result = resolveStyles(component, renderedElement);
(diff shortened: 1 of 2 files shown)


modules/resolve-styles.js
```diff
@@ -137,6 +137,14 @@ var resolveStyles = function (
   // will not be here, so each component needs to use Radium.
   var newChildren = null;
   var oldChildren = renderedElement.props.children;
+  var resolveChild = function (child) {
+    // test if child is a valid ReactDOMElement and not ReactComponentElement
+    if (React.isValidElement(child) && typeof child.type === 'string') {
```
ianobermiller: Instead of the extra closure, which I find a bit hard to follow, let's just invert this check and return early from resolveStyles directly after the children resolution:

```js
if (!React.isValidElement(renderedElement) || typeof renderedElement.type !== 'string') {
  return renderedElement;
}
```

The check must come after child resolution because you could pass dom elements as children of custom components, and you'd want them to get resolved correctly. Please add a test like the following: 

```as
var renderedElement = (
  <div>
    <CustomComponent style={style}>
      <div style={style} />
    </CustomComponent>
  </div>
);
```

Also, once you add the check, you can remove it from inside the callback in React.Children.map.

modules/__tests__/resolve-styles-test.js
```diff
@@ -761,4 +761,18 @@ describe('resolveStyles', function () {
     });
   });
 
+  describe('ReactComponentElement children', function () {
+    it('doesn\'t change ReactComponentElement children', function () {
+      var component = genComponent();
+      class CustomComponent extends React.Component {}
+      var style = {':hover': {}};
+      var renderedElement = <div>
```
ianobermiller: Please wrap multiline JSX in parens.

azazdeaz:
@ianobermiller You're right! I added the changes, what do you think?

modules/resolve-styles.js
```diff
@@ -161,6 +161,20 @@ var resolveStyles = function (
...
+  if (
+    !React.isValidElement(renderedElement) ||
+    typeof renderedElement.type !== 'string'
+  ) {
+    if (oldChildren === newChildren) {
+      return renderedElement;
+    }
+
+    return React.cloneElement(
+      renderedElement, renderedElement.props, newChildren
```
ianobermiller: if it doesn't fit on one line in 80 chars or so, drop each param to its own line

ianobermiller:
We can clean that up another time, I'll probably make a refactoring pass at resolveStyles in the future.

PR #208, opened 2015-06-14
Prefixer fixes for server rendering

### LINKED ISSUE #207: Different rendering camelCase key on server and client
Why this code now renders different on server and client

`const rules = {
  body: {
    margin: 0,
    fontFamily: Theme.fontFamily,
    fontWeight: 300,
    fontSize: Theme.px.fontSize
  }
}`

from console 
`Warning: React attempted to reuse markup in a container but the checksum was invalid. This generally means that you are using server rendering and the markup generated on the server was not what the client was expecting. React injected new markup to compensate which works but you have lost many of the benefits of server rendering. Instead, figure out why the markup being generated is different on the client or server:
 (client) body{margin: 0;
font-family: 'Exo 2', sa
 (server) body{margin: 0;
fontFamily: 'Exo 2', san
`
difference in that on server it renders `fontFamily` and on client `font-family`

Before i upgraded to version 0.13.0 i use 0.12.0 and it was working well

ianobermiller:
- Fixes #207, not using dash-case on server with Style
- Fixes another latent bug, when using a fallback array as value, the
server should pick the first one (not perfect, since it will mean older
browsers will be more likely to have to throw out state)

### DIFF
--- a/modules/__tests__/prefixer-test.js
+++ b/modules/__tests__/prefixer-test.js
@@ -62,7 +62,7 @@ describe('Prefixer', () => {
     expect(Prefixer.jsPrefix).toBe('Webkit');
   });
 
-  it('doesn\'t detect prefix if not in browser', () => {
+  it('doesn\'t detect prefix if in server', () => {
     jest.setMock('exenv', {
       canUseDOM: false,
       canUseEventListeners: false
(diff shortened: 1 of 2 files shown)


alexlande:
:+1:

PR #209, opened 2015-06-14
Update dependencies, lint

ianobermiller:
- Include react linter plugin
- remove use strict, babel adds it automatically for modules


alexlande:
:ship:

PR #218, opened 2015-06-20
Warn when mixing longhand and shorthand props

### LINKED ISSUE #95: Decompose shorthand/compound styles to their constituent parts
I have some buttons in a nav with a default padding value:

```
var style = {
  padding: '10px 20px'
}
```

When selected, I add a thin left border + adjust only the left padding accordingly:

```
modifiers: [
  {
    selected: {
      borderLeft: '3px solid blue',
      paddingLeft: 17
    }
  }
]
```

Unfortunately, when the item becomes unselected, Radium is using `0` for the `paddingLeft` value instead of the original `20px` value. I’m assuming this is because I don’t have `paddingLeft` as a property in the default styles:

I was able to fix this by changing `paddingLeft` to `padding` + using a more explicit string value:

```
modifiers: [
  {
    selected: {
      borderLeft: '3px solid blue',
      padding: '10px 20px 10px 17px'
    }
  }
]
```

My gut tells me Radium should be able to intelligently merge related properties like `padding` + `paddingLeft`, `margin` + `marginTop`, etc. on modifier changes.

1. Is this the expected behavior?
2. Is it recommended to be explicit with value changes like these?

ianobermiller:
Fixes #95

### DIFF
--- a/modules/__tests__/resolve-styles-test.js
+++ b/modules/__tests__/resolve-styles-test.js
@@ -823,4 +823,52 @@ describe('resolveStyles', function () {
       });
     });
   });
+
+  /* eslint-disable no-console */
+  describe('warnings', function () {
+    beforeEach(function () {
+      this.originalConsoleWarning = console.warning;
+      console.warning = jest.genMockFunction();
+    });
+
+    afterEach(function () {
+      console.warning = this.originalConsoleWarning;
+      process.env.NODE_ENV = null;
+    });
+
+    it('warns when mixing longhand and shorthand properties', function () {
+      var component = genComponent();
+      var renderedElement = (
+        <div style={{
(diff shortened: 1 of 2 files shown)


alexlande:
Nice, looks good. You've got a little lint though.

williamboman:
Quite annoying to see these warnings when it's done deliberately, e.g.;

```js
// utils.radium.js
export const resetBoxModel = { margin: 0, padding: 0, border: 0 };

// foo.jsx
import {resetBoxModel} from './utils.radium';

const styles = {
    base: {
        ...resetBoxModel,
        marginRight: 5,
        paddingLeft: 10,
    }
};
```

or am I doing something wrong here?

PR #223, opened 2015-06-23
Fix fallback for numeric values

### LINKED ISSUE #221: Fallback expressions don't work when first is numeric
I'm trying to calculate height of an element based on the viewport height using `vh` units and fallback to the actual viewport height I'm injecting into the component as `height` property:

```javascript
    let backgroundStyle = {
      height: [`${height - headerHeight}px`, `calc(100vh - ${headerHeight}px)`]
    }
```

However, my component gets rendered with the following style
```css
height: 343px,calc(100vh - 64px)
```

packages:
```javascript
    "react": "0.13.3",
    "radium": "0.13.0"
```

I'm getting this CSS both on the server and on the client.

Interestingly enough, if I switch the order of the values:

```javascript
height: [`calc(100vh - ${headerHeight}px)`, `${height - headerHeight}px`]
```

I'm only getting the following CSS:

```css
height: calc(100vh - 64px)
```

ianobermiller:
Fixes #221, fallback when first value is a number. Also needs to add px
to non-unitless properties so the value checks can pass, otherwise
`[‘calc(…)’, 400]` would end up with calc, even if calc wasn’t
supported.

### DIFF
--- a/modules/prefixer.js
+++ b/modules/prefixer.js
@@ -53,6 +53,40 @@ var infoByCssPrefix = {
   }
 };
 
+/**
+ * CSS properties which accept numbers but are not in units of "px".
+ * Copied from React core June 22, 2015.
+ * https://github.com/facebook/react/blob/
+ * ba81b60ad8e93b747be42a03b797065932c49c96/
+ * src/renderers/dom/shared/CSSProperty.js
+ */
+var isUnitlessNumber = {
+  boxFlex: true,
+  boxFlexGroup: true,
+  columnCount: true,
+  flex: true,
+  flexGrow: true,
+  flexPositive: true,
+  flexShrink: true,
+  flexNegative: true,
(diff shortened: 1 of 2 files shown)


modules/prefixer.js
```diff
@@ -127,13 +161,36 @@ var getPrefixedPropertyName = function (property) {
   return prefixedPropertyCache[property] = workingProperty;
 };
 
-var _getPrefixedValue = function (property, value, originalProperty) {
-  // don't test numbers or numbers with units (e.g. 10em)
+// React is planning to deprecate adding px automatically
+// (https://github.com/facebook/react/issues/1873), and if they do, this
```
alexlande: Wasn't aware of that. Cool.

alexlande:
Little linty. :shipit:

PR #229, opened 2015-06-25
Fixes #228: Fix warning when mixing longhand and shorthand props

### LINKED ISSUE #228: Use console.warn instead of console.warning
In PR https://github.com/FormidableLabs/radium/pull/218, `console.warning()` is used several times for logging an error.

Shouldn't it be calling `console.warn()`? Safari is throwing an error:

```
undefined is not a function (evaluating 'console.warning')
```

angeloashmore:
`console.warning` was called rather than `console.warn` in the original PR.


ianobermiller:
Build failure due to a bug in babel 5.6.8, fixed in [5.6.10](https://github.com/babel/babel/blob/master/CHANGELOG.md#5610).

PR #230, opened 2015-06-26
Changed console.warning (not a function) to console.warn

gragland:
Otherwise you get: Uncaught TypeError: console.warning is not a function


gragland:
Closing because there is already a pull request for this.

PR #237, opened 2015-07-02
Stop mutating style state

### LINKED ISSUE #236: Implementing shouldComponentUpdate with Radium states
I'm trying to implement `shouldComponentUpdate` for my component which uses Radium for `active` and `hover` states.  I'm running into a problem where it looks like Radium is updating the current state and next state at the same time so that you can't determine if the value has changed which prevents the component from rendering.

````
shouldComponentUpdate(nextProps, nextState) {
  console.log('Current state: ', Radium.getState(this.state, 'btn', ':hover'));
  console.log('Next state: ', Radium.getState(nextState, 'btn', ':hover'));

  return 
    Radium.getState(nextState, 'btn', ':hover') !== Radium.getState(this.state, 'btn', ':hover');
},
````

The first hover in works as expected.  The current state is false and the next state is true.

````
Current state:  false
Next state:  true
````

After that, when you leave the element, the current state is already false even though it had just been set to true on the last state change.  Hovering over and out of the element now has no effect since the states are always set to the same value.

````
Current state:  false
Next state:  false

Current state:  true
Next state:  true

Current state:  false
Next state:  false

````

ianobermiller:
Fixes #236

### DIFF
--- a/modules/__tests__/resolve-styles-test.js
+++ b/modules/__tests__/resolve-styles-test.js
@@ -356,6 +356,29 @@ describe('resolveStyles', function () {
         result = resolveStyles(component, renderedElement);
         expect(result.props.style.background).toEqual('blue');
       });
+
+      it('doesn\'t mutate state', function () {
+        var component = genComponent();
+        var style = {background: 'blue'};
+        style[':' + pseudo] = {background: 'red'};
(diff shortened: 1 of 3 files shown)


ianobermiller:
Windows failure is unrelated.

PR #238, opened 2015-07-02
Don't alias window.matchMedia

### LINKED ISSUE #234: matchMedia export breaks IE11
Pull request #146 broke IE11. The matchMedia export overrides the built-in matchMedia function which causes IE to throw an Invalid calling object exception and will not load the page.

ianobermiller:
Instead, wrap it in a function. This *should* fix #234.


bunkat:
Just wanted to confirm that this fix corrected the issue I was seeing.  Thanks!

PR #240, opened 2015-07-03
Migrate to Karma for testing

### LINKED ISSUE #225: Move tests from jest to karma
Jest just hasn't been moving very quickly development-wise. I wonder if we should move our tests to karma. We can use `karma-commonjs-plus` for mocking and `isparta-loader` to get proper code coverage reports again. Also, our tiny test suite takes ~2.5 seconds to run, which seems like way too long.

exogen:
- Translate Jest assertions and spies to Chai/Sinon.
- Use `inject-loader` for mocking `require()`'d dependencies. I tried both `karma-commonjs-plus` and `mockery` and couldn't get either of them to work after much debugging and fiddling around. `inject-loader`'s API is quite unwieldy, but it's the only one that worked.
- Fixes up the flow type annotation syntax in two modules; the trailing commas were causing an error to be thrown during transformation, not sure why.
- Remove all references to Jest
- Add Node 0.12 back to the Travis config
- I made a point to keep all the `babel` dependency versions the same, so I'm not sure why the `dist` files changed slightly.

fixes #225


webpack.config.js
```diff
@@ -23,7 +23,7 @@ module.exports = {
   module: {
     loaders: [
       {
-        test: /\.js(x|)?$/,
+        test: /\.js(x|)$/,
```
ianobermiller: Would `/\.jsx?$/` work and be cleaner?

modules/__tests__/prefixer-test.js
```diff
@@ -1,119 +1,113 @@
...
-    var Prefixer = require('../prefixer.js');
+    var Prefixer = require('inject!prefixer.js')({'exenv': exenv});
     var result = Prefixer.getPrefixedStyle({display: 'flex'});
 
-    expect(result).toEqual({display: 'flex'});
-    expect(Prefixer.cssPrefix).toBe('');
-    expect(Prefixer.jsPrefix).toBe('');
-    expect(document.createElement).not.toBeCalled();
-    expect(window.getComputedStyle).not.toBeCalled();
+    expect(result).to.eql({display: 'flex'});
```
ianobermiller: Please use he full equal version everywhere

ryan-roemer: Don't we want a `.deep.equal` (aka the alias `.eql`) here?  Strict equality wouldn't match.

exogen: @ryan-roemer is right, `.equal` would fail for these objects. @ianobermiller, are you saying you prefer the `.deep.equal` syntax to `.eql`?

webpack.config.js
exogen: Yes! I was tempted to change it to that – actually I didn't want to modify this file at all to keep the diff smaller; this change is a mistake. But I'll change it to that so the syntax matches what's in `karma.conf.js`.

The `loader` here is bugging me too; in the Karma config it's `babel?stage=1` and here it's `babel-loader?stage=0`. The former syntax is what I'm seeing in recommended in documentation everywhere, do you know why this is different?

ianobermiller:
Wow nicely done! Please revert the dist files, we only generate them when
tagging releases.

ryan-roemer:
Appveyor's failing (https://ci.appveyor.com/project/ryan-roemer/radium/build/1.0.118) and this seems familiar. Probably related, Karma has historically been wonky with installing internal deps on practically all CI systems.

We can try kicking Appveyor, but in addition we can consider a few extra hacks I have here for a Karma CI build https://github.com/FormidableLabs/full-stack-testing/blob/master/appveyor.yml

ryan-roemer:
OK. I've kicked Appveyor. Let's watch build https://ci.appveyor.com/project/ryan-roemer/radium/build/1.0.119

modules/__tests__/prefixer-test.js
```diff
@@ -1,119 +1,113 @@
...
+    window.getComputedStyle.restore();
   });
 
   it('detects Firefox prefix', () => {
     browserPrefix = '-moz-';
-    var Prefixer = require('../prefixer.js');
-    expect(Prefixer.cssPrefix).toBe('-moz-');
-    expect(Prefixer.jsPrefix).toBe('Moz');
+    var Prefixer = require('inject!prefixer.js')({'exenv': exenv});
+    expect(Prefixer.cssPrefix).to.equal('-moz-');
```
ryan-roemer: Future comment: You'll get more informative assertion failures with something like:

```js
expect(Prefixer).to.have.property('cssPrefix', '-moz-');
```

... but I don't want to clog up a straight porting PR with change suggestions ;)

modules/__tests__/resolve-styles-test.js
```diff
@@ -395,10 +391,10 @@ describe('resolveStyles', function () {
         result.props[onHandlerName]();
         result.props[offHandlerName]();
 
-        expect(originalOffHandler).toBeCalled();
+        expect(originalOffHandler).to.have.been.called;
 
         result = resolveStyles(component, renderedElement);
-        expect(result.props.style.background).toEqual('blue');
+        expect(result.props.style.background).to.equal('blue');
```
ryan-roemer: And again, for the future, would get much better assert failures here with something like:

```js
expect(result.prop).to.have.deep.property('style.background', 'blue');
```

ryan-roemer:
@exogen -- Great work! And nice to see `v0.12` enabled on both Travis and AppVeyor.

I'll leave you to @ianobermiller and @alexlande for the rest of the substantive approval here...

exogen:
Thanks for the assertion tips, @ryan-roemer – I'll open a separate issue to track those once this PR is merged.

modules/__tests__/prefixer-test.js
ianobermiller: ah yes, please change `.eql` to `.deep.equal`, much more readable

.travis.yml
```diff
@@ -2,10 +2,7 @@ language: node_js
 
 node_js:
   - 0.10
-  # TODO: Re-enable v0.12 builds.
-  # Radium Ticket: https://github.com/FormidableLabs/radium/issues/138
-  # Upstream Jest Ticket: https://github.com/facebook/jest/issues/243
-  #- 0.12
+  - 0.12
```
ianobermiller: should we add an io.js build in here as well?

karma.conf.js
```diff
@@ -0,0 +1,79 @@
...
+        }],
+        loaders: [{
+          test: /\.jsx?$/,
+          exclude: [/node_modules/],
+          loader: 'babel?stage=1'
+        }, {
+          test: /\.css$/,
+          loader: 'style-loader!css-loader'
+        }, {
+          test: /\.styl$/,
```
ianobermiller: remove stylus code

exogen:
Added "iojs" to the Travis config, which will pick up the latest stable release. For Appveyor, the docs say this will pick up io.js: `  - nodejs_version: "1.0"` which seems a bit sketch to me (just the version number differentiating the runtime). Should we add it there too anyway?

ianobermiller:
I don't see why not. Can always change it back if there are issues.

ianobermiller:
Awesome work @exogen, thank you!

PR #241, opened 2015-07-05
ensuring that defaultProps get transferred in every browser

### LINKED ISSUE #235: Radium breaks getDefaultProps on IE
I've tested this on IE9/10 (internet is too slow to download an 11 VM right now) and created some examples on my fork that demonstrate that it is Radium that is responsible.

https://github.com/bobbyrenwick/radium/tree/test-out-ie

I can't see anything that makes it immediately obvious why this is happening

bobbyrenwick:
This fixes #235, and doesn't require a move away from using the subclass method in order to enhance components.

### DIFF
--- a/modules/enhancer.js
+++ b/modules/enhancer.js
@@ -44,6 +44,10 @@ var enhanceWithRadium = function (ComposedComponent: constructor): constructor {
     }
   }
 
+  if (ComposedComponent.hasOwnProperty('defaultProps')) {
+    RadiumEnhancer.defaultProps = ComposedComponent.defaultProps;
+  }
+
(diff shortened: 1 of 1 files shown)


ianobermiller:
Don't we need to do the same for `propTypes`, `contextTypes`, and `childContextTypes`?

bobbyrenwick:
@ianobermiller good thinking, that's done too

ianobermiller:
Can you just add a comment why we are doing this? Just say this is for IE <10 and link to the babel caveats page. Then this should be good to go. Thanks again!

bobbyrenwick:
@ianobermiller done!

ianobermiller:
Nice work, thanks @bobbyrenwick!

PR #242, opened 2015-07-05
Remove peerDependencies so we can use it with react 0.14.0-beta1

dariocravero:


ianobermiller:
Following [react-router](https://github.com/rackt/react-router/blob/master/package.json), let's move it to devDependencies and change to `"react": "0.13.x",` so the examples don't break.

dariocravero:
@ianobermiller sorry I missed your comment, `react` [is already in devDependencies](https://github.com/FormidableLabs/radium/blob/master/package.json#L60) although locked to `>=0.13.3`.

ianobermiller:
Thanks, I missed that.

dariocravero:
Class :) Could you do a new minor release so that we can use the npm package directly? Thanks!

ianobermiller:
Yeah I just want to merge #250 then I'll cut a release.

dariocravero:
Perfect, thanks! :)

PR #250, opened 2015-07-09
Warn & ignore when receiving null or undefined values

### LINKED ISSUE #245: Warn on and ignore null or undefined values
I made a stackoverflow outlining the issue:

http://stackoverflow.com/questions/31264376/reactjs-radium-altjs-uncaught-typeerror-cannot-read-property-tostring-of

AnSavvides:
Title says it all!

I also added a couple of tests and made a change to `.eslintrc` to allow the use of `undefined` since we need to compare against it & use it for testing.

Closes #245.

### DIFF
--- a/modules/prefixer.js
+++ b/modules/prefixer.js
@@ -191,6 +191,7 @@ var _getPrefixedValue = function (property, value, originalProperty) {
         if (console && console.warn) {
           console.warn('CSS value is "' + value + '" for property "' + property + '"');
         }
+        return value;
       }
     }
 
--- a/modules/prefixer.js
+++ b/modules/prefixer.js
@@ -184,7 +184,14 @@ var _getPrefixedValue = function (property, value, originalProperty) {
     }
 
     if (typeof value !== 'string') {
-      value = value.toString();
+      if (value !== null && value !== undefined) {
+        value = value.toString();
+      } else {
+        /* eslint-disable no-console */
+        if (console && console.warn) {
+          console.warn('CSS value is "' + value + '" for property "' + property + '"');
+        }
+      }
     }
 
     // don't test numbers with units (e.g. 10em)
--- a/.eslintrc
+++ b/.eslintrc
@@ -147,7 +147,7 @@ rules:
     no-undef: 2                    # disallow use of undeclared variables unless mentioned in a
                                    #    /*global */ block
     no-undef-init: 2               # disallow use of undefined when initializing variables
-    no-undefined: 2                # disallow use of undefined variable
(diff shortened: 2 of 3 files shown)


modules/prefixer.js
```diff
@@ -184,7 +184,15 @@ var _getPrefixedValue = function (property, value, originalProperty) {
     }
 
     if (typeof value !== 'string') {
-      value = value.toString();
+      if (value !== null && value !== undefined) {
+        value = value.toString();
+      } else {
+        /* eslint-disable no-console */
+        if (console && console.warn) {
+          console.warn('CSS value is "' + value + '" for property "' + property + '"');
```
ianobermiller: Can you please add the component name as in #253?

AnSavvides: I think I might be missing something, but we don't have a reference to a `component` within `prefixer`? Should `getPrefixedStyle` be taking a `component` as an argument for the sake of adding the name in the warning?

ianobermiller:
Yeah true, that would be crummy. I also wonder if we should just use `typeof value === 'undefined'` instead of removing the lint rule, but I don't care too much, so merge it is!

bobbyrenwick:
@ianobermiller @AnSavvides I think it's really useful to have the component name when you're debugging these kinds of errors, as well as unsupported CSS values / properties.

What about moving the responsibility of `console.warn` to the caller of `getPrefixedStyle` and having `getPrefixedStyle` return errors to the caller in some way? That way we would know what component we were rendering at the time.

AnSavvides:
@bobbyrenwick if we really wanted to have the component name in the warning, I would favor just passing `component` as an argument over having the warning in the caller of `getPrefixedStyle` - I just think that's a little cleaner, maybe? Not too fussed in any case!

@ianobermiller if you think that's a good idea (either approach) feel free to create an issue, assign it to me and I'll happily sort it out :)

ianobermiller:
Good discussion. Let's pass the component name to the prefixer just to make the warnings better. Also, thoughts on enabling warnings only in DEV?

AnSavvides:
I would be in favor of enabling warnings only in DEV, I think that's good practice. If it makes any difference, that's how it's done in React too ([here](https://github.com/facebook/react/blob/ba81b60ad8e93b747be42a03b797065932c49c96/src/renderers/shared/event/eventPlugins/ResponderTouchHistoryStore.js#L105-L107) for example), so React developers should be used to only getting warnings in DEV.

bobbyrenwick:
I'm also a fan of only having them in dev. We can use the same approach as I did for the hot loading fix.

PR #252, opened 2015-07-10
extending prototype functions in order to fix hot loader

### LINKED ISSUE #219: 0.13.0 breaks react-hot-loader
I'm using `React.createClass`, `Radium` and `react-hot-loader@1.2.7` like follows:

```es6
var React = require("react");
var Radium = require("radium");

var SearchInput = React.createClass({

    getStyles: function() {
        const styles = {
            base: {
                backgroundColor: "red",
            },
        };

        return [
            styles.base,
        ];
    },

    render: function() {
        return (
            <div style={this.getStyles()}>
                Hello Radium
            </div>
        );
    }

});

module.exports = Radium(SearchInput);
```

When I updated to `v0.13.0`, hot updates to components using Radium stopped working. It was working with `v0.12.1` before. This is a pretty big as all my styles are done with Radium and this now stops me from seeing the effect of my code changes.

bobbyrenwick:
This also removes the setting up of the initial `this.state._radiumStyleState` because there's no way to reconcile the differences between ES6 classes and React.createClass without creating a wrapping constructor that kills hot loading. It also removes the tests for that behaviour because it doesn't actually seem to negatively affect the behaviour.

You can see that this branch works with hot loader by cloning my fork of `react-hot-boilerplate` - https://github.com/bobbyrenwick/react-hot-boilerplate/tree/test-radium-hoc.

The only thing I'm not too sure about is that I've deleted some Flow annotations and I'm not sure what sort of effects that will have as I haven't had much time to play around with it! @ianobermiller is there anything Flow related that needs to be done?

Fixes #219

### DIFF
--- a/modules/__tests__/enhancer-test.js
+++ b/modules/__tests__/enhancer-test.js
@@ -6,32 +6,6 @@ var Enhancer = require('inject!enhancer.js')({
 var {Component} = require('react');
 
 describe('Enhancer', () => {
-  it('sets up initial state', () => {
-    class Composed extends Component { }
-    var Enhanced = Enhancer(Composed);
-
-    var instance = new Enhanced();
-
-    expect(instance.state).to.deep.equal({_radiumStyleState: {}});
-  });
-
-  it('merges with existing state', () => {
-    class Composed extends Component {
-      constructor () {
-        super();
-        this.state = {foo: 'bar'};
-      }
-    }
-    var Enhanced = Enhancer(Composed);
-
-    var instance = new Enhanced();
-
-
-    expect(instance.state).to.deep.equal(
-      {foo: 'bar', _radiumStyleState: {}}
-    );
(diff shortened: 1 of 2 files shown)


bobbyrenwick:
So I've been testing this out on our app today and it seems that there are some issues with this. I will let you know more once I've done some more testing.

bobbyrenwick:
So the issue was caused by removing, `_radiumStyleState` from the `initialState` while using `PureRenderMixin`. For example, on a hover, in `shallowEqual` we end up comparing a state of `null` (the initial state) to a state of `{"_radiumStyleStyle": {...etc}}` and get `cannot read property hasOwnProperty of null`

bobbyrenwick:
Closing in favour of #255 because that doesn't have the same issues with `PureRenderMixin` and no initial state being set.

Issue #259, opened 2015-07-14
Pass component name to prefixer for better warnings


ianobermiller:
I'll leave it to your discretion where you add this (last/first argument).

cc @AnSavvides, who offered to take this on!

AnSavvides:
:+1:

PR #263, opened 2015-07-14
Allow falsy style values

clessg:
The ability to do `<Component style={{ height: null }} />` was broken in #95. That behavior is useful in a few situations and this pull request should restore it.

### DIFF
--- a/modules/__tests__/resolve-styles-test.js
+++ b/modules/__tests__/resolve-styles-test.js
@@ -902,7 +902,7 @@ describe('resolveStyles', function () {
       expect(console.warn).to.not.have.been.called;
     });
 
-    it('does not warn when passed a falsy entry value', function () {
+    it('does not throw when passed a falsy entry value', function () {
       var component = genComponent();
       var renderedElement = (
         <div style={{
--- a/modules/__tests__/resolve-styles-test.js
+++ b/modules/__tests__/resolve-styles-test.js
@@ -901,6 +901,19 @@ describe('resolveStyles', function () {
 
       expect(console.warn).to.not.have.been.called;
     });
+
+    it('does not warn when passed a falsy entry value', function () {
+      var component = genComponent();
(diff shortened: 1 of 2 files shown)


ianobermiller:
Does `{height: null}` do what you'd expect, unset the height style? Does `undefined` do the same as well? Sounds like we want to fix it here and also remove the warning in the prefixer that was added in #250.

ianobermiller:
You'll probably also want to update the prefixer, which will throw warnings. I suggest we only allow `=== null`, to make sure it is intentional and not a bug.

AnSavvides:
@ianobermiller where does that leave us with #259? I was getting close to opening a PR, pending some test updates.

ianobermiller:
@AnSavvides send out a PR for #259, this one still needs some work

scottcorgan:
Any idea when this will get merged? Sent me debugging for awhile today to see what it was.

clessg:
@ianobermiller My use for this particular change is to allow for a value to be conditionally set if it isn't `undefined`. Assume we have a component with a `height` prop:

    const { height, x, y } = this.props;
    const imageStyle = {
      height,
      marginLeft: -x,
      marginTop: -y
    };

    return ;

In a previous version of Radium, if no `height` prop was provided, then the `height` would not be put into the resulting markup. Other ways of achieving this:
* `if (height === void 0) { delete imageStyle.height; }`
* `if (height !== void 0) { imageStyle.height = height; }`.
* Set `defaultProps.height` to `'initial'`.

In other words, not a big deal.

This bug was really hard to track down though, so at least having an error message or something would be great, even if it doesn't allow for this particular pattern. Regarding your other question, `style={[{ height: 100 }, { height: null }]}` results in a `height` of 100.

scottcorgan:
Any way we can get a patch release on NPM for this issue? Thanks.

Issue #265, opened 2015-07-15
function as children seems to break when using Radium


Cottin:
The pattern of having a function as children as seen e.g. here:
https://twitter.com/dan_abramov/status/605891462069248000
https://discuss.reactjs.org/t/children-as-a-function-render-callbacks/626
https://github.com/gaearon/redux

...seems to break when using Radium.

Works:
````
MyChild = React.createFactory React.createClass
  propTypes:
    children: React.PropTypes.func
  render: ->
   div {}, this.props.children()

MyParrent = React.createClass
  render: ->
    MyChild null, () -> 1

module.exports = MyParrent
````

Doesen't work:
````
MyChild = React.createClass
  propTypes:
    children: React.PropTypes.func
  render: ->
    div {}, this.props.children()

MyParrent = React.createClass
  render: ->
    MyChild null, () -> 1

module.exports = Radium MyParrent     <------ difference
````

`children` gets evaluated to some fragment object instead of a function:
````
console> this.props.children
Object {_reactFragment: Object, _reactDidWarn: false}
````

Is this a known issue or am I messing something up?

PR #266, opened 2015-07-15
add support for "function as children" pattern

Cottin:


Cottin:
refs #265

ianobermiller:
Looks good. Downside to using a function as children is that you cannot use Radium on the stuff rendered by those children (unless it is wrapped in another component, of course). We may want to add something like this to the FAQ.

Cottin:
ohh, damnit I didn't think that far :)  That would be kind of a deal breaker for me at least.

PR #267, opened 2015-07-15
Rewrite prefixMatch to use a for loop instead of Array prototype

bsbeeks:
This is an extreme use case, but one that we encountered while testing our web app inside an Android web view (chromium).

Calling `Array.prototype.slice` with the `windowStyles` object causes an `Uncaught TypeError: undefined is not a function` in older versions Android (KitKat) web views. This rewrites it to use a for loop.

- `npm run lint` passed
- `npm test` passed
- `flow` returned a few errors in files unrelated to this PR

**Note:** I couldn't figure out *why* it was returning that error, but did confirm that it was that line causing the error. We monkey patched the module with this fix and it eliminated the error. Perhaps older versions of chromium don't support calling an Array prototype method with an Object?

### DIFF
--- a/modules/prefixer.js
+++ b/modules/prefixer.js
@@ -100,10 +100,17 @@ if (ExecutionEnvironment.canUseDOM) {
 
   // Based on http://davidwalsh.name/vendor-prefix
   var windowStyles = window.getComputedStyle(document.documentElement, '');
-  var prefixMatch = Array.prototype.slice.call(windowStyles)
-    .join('')
-    .match(/-(moz|webkit|ms|o)-/);
-  var cssVendorPrefix = prefixMatch && prefixMatch[0];
+  var prefixMatch, cssVendorPrefix;
+
+  for (var i = 0; i < windowStyles.length; i++) {
+    prefixMatch = windowStyles[i].match(/-(moz|webkit|ms|o)-/);
+
+    if (prefixMatch) {
+      break;
+    }
+  }
+
+  cssVendorPrefix = prefixMatch && prefixMatch[0];
 
   prefixInfo = infoByCssPrefix[cssVendorPrefix] || prefixInfo;
 }
--- a/modules/prefixer.js
+++ b/modules/prefixer.js
@@ -103,6 +103,9 @@ if (ExecutionEnvironment.canUseDOM) {
   var prefixMatch;
   var windowStyles = window.getComputedStyle(document.documentElement, '');
 
+  // Array.prototype.slice.call(windowStyles) fails with
+  // "Uncaught TypeError: undefined is not a function"
+  // in older versions Android (KitKat) web views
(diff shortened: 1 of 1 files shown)


modules/prefixer.js
```diff
@@ -100,10 +100,17 @@ if (ExecutionEnvironment.canUseDOM) {
 
   // Based on http://davidwalsh.name/vendor-prefix
   var windowStyles = window.getComputedStyle(document.documentElement, '');
-  var prefixMatch = Array.prototype.slice.call(windowStyles)
-    .join('')
-    .match(/-(moz|webkit|ms|o)-/);
-  var cssVendorPrefix = prefixMatch && prefixMatch[0];
+  var prefixMatch, cssVendorPrefix;
```
ianobermiller: split var declarations onto separate lines

modules/prefixer.js
```diff
@@ -100,10 +100,17 @@ if (ExecutionEnvironment.canUseDOM) {
...
+
+  for (var i = 0; i < windowStyles.length; i++) {
```
ianobermiller: Add a comment about why we made this ugly ;)

```
// Array.prototype.slice.call(windowStyles) fails with "Uncaught TypeError: undefined is not a function" in older versions Android (KitKat) web views

**Comment by ianobermiller** [comment]
Thanks for trying out Radium in a non-mainstream browser!

**Comment by bsbeeks** [comment]
@ianobermiller updated with your suggested changes

**Comment by bsbeeks** [comment]
Can't seem to find the ESLint error that Travis is failing on. Line 217 of `modules/prefixer.js` looks correct.

File: modules/prefixer.js [hunk 3]
```diff
@@ -99,11 +99,22 @@ if (ExecutionEnvironment.canUseDOM) {
...
-  var prefixMatch = Array.prototype.slice.call(windowStyles)
-    .join('')
-    .match(/-(moz|webkit|ms|o)-/);
-  var cssVendorPrefix = prefixMatch && prefixMatch[0];
+
+  // Array.prototype.slice.call(windowStyles) fails with
+  // "Uncaught TypeError: undefined is not a function"
+  // in older versions Android (KitKat) web views
+  for (var i = 0; i < windowStyles.length; i++) {
+    prefixMatch = windowStyles[i].match(/-(moz|webkit|ms|o)-/);
```
ianobermiller: One more thing, can you pull the regex here out into a constant, like `VENDOR_PREFIX_REGEX`

bsbeeks: No problem. Where do you prefer constants to be defined? Top of the file, top of method, constant's file, etc?

ianobermiller: Top of file is fine, just so that the regex is only instantiated once for the entire module

bsbeeks:
prefix regex moved to variable at top of file. Let me know if there is anything else.

PR #277, opened 2015-07-20
fix older firefox missing float property problem

bencao:
For older Firefox(for our case, they're version 30 and 31), the element created by document will have no 'float' key for its style.
This caused malfunction in the UI because all float attributes were missing.

If this is the only case, then we propose a simple fix by adding a float attribute to domStyle.
Please correct me if there're better ways.


ianobermiller:
Well that is odd. I wonder how many other bugs are lurking due to this?

PR #279, opened 2015-07-20
Better support for old and tweener flexbox syntax 

### LINKED ISSUE #205: -ms-flex-pack
IE10 doesn't support justify-content...it needs "-ms-flex-pack". I can put in "-ms-flex-pack" in my styles specifically, but it tells me it's an unsupported CSS property in other browsers (which makes sense). Any ideas how to fix?

sylvaingi:
There are 2 main caveats with supporting the 2009 'old' syntax, first an element with `display: box` is treated as an inline element, and `box-flex` accepts only argument instead of 3 in the latest spec.

Fixes #205

### DIFF
--- a/modules/prefixer.js
+++ b/modules/prefixer.js
@@ -15,12 +15,28 @@ var infoByCssPrefix = {
     alternativeProperties: {
       // OLD - Firefox 19-
       flex: [{css: '-moz-box-flex', js: 'MozBoxFlex'}],
-      order: [{css: '-moz-box-ordinal-group', js: 'MozBoxOrdinalGroup'}]
+      order: [{css: '-moz-box-ordinal-group', js: 'MozBoxOrdinalGroup'}],
+      flexDirection: [{css: '-moz-box-orient', js: 'MozBoxOrient'}],
+      alignItems: [{ css: '-moz-box-align', js: 'MozBoxAlign' }],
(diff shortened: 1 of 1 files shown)


modules/prefixer.js
```diff
@@ -15,23 +15,67 @@ var infoByCssPrefix = {
     alternativeProperties: {
       // OLD - Firefox 19-
       flex: [{css: '-moz-box-flex', js: 'MozBoxFlex'}],
-      order: [{css: '-moz-box-ordinal-group', js: 'MozBoxOrdinalGroup'}]
+      order: [{css: '-moz-box-ordinal-group', js: 'MozBoxOrdinalGroup'}],
+      flexDirection: [{css: '-moz-box-orient', js: 'MozBoxOrient'}],
+      alignItems: [{ css: '-moz-box-align', js: 'MozBoxAlign' }],
```
ianobermiller: nit: sort this list alphabetically, and remove spaces inside the braces

ianobermiller:
Awesome work, thank you! Just a little clean up and this will be good to merge. I've also fixed the error on master so travis should be passing.

sylvaingi:
Cleaned up the code. Also I've not mentioned that it's not tested on old webkit, although on Firefox 19 it renders alright.

skevy:
Woo!

Issue #285, opened 2015-07-21
Tests are slow


ianobermiller:
The tests as we have them setup through karma are quite slow, roughly 10 seconds on my 2015 MBP. I'm assuming during development we want to have (or already have?) a faster method than `npm test`

cc @exogen

ryan-roemer:
@ianobermiller -- Perhaps first thing to do is a baseline with all tests skipped. There's going to be some overhead from moving to real browsers / phantom over JsDom (which jest uses) but my hunch is we've amortized it at that 10 seconds and it's not linear increases from here.

Additional tests should be as fast as anything else. We can then pair this with additional work to have the karma server already initialized and watching for a common case "ongoing development" workflow.

exogen:
I think during development, we could have a command like `test-dev` that doesn't use Karma's `singleRun` option (which I think is there for better CI behavior). That way, the Karma server is only started up once, and re-running tests would ideally be instantaneous. But I haven't tried it yet.

ianobermiller:
@exogen I was playing around with it and that seemed to be correct.

Since we are mocking browser APIs anyway, why not just get rid of Karma and run the tests in Node? By default? We could still run them in other browsers during CI using a package like browser-run. I may be slightly influenced by just having read this post: https://medium.com/javascript-scene/why-i-use-tape-instead-of-mocha-so-should-you-6aa105d8eaf4

exogen:
@ianobermiller I tried this as a quick test. React does some DOM stuff internally that we aren't mocking right now. I got as far as most of the tests running in pure Node, but there are errors like:

```
/Users/brianbeck/Projects/Formidable/radium/node_modules/react/lib/CSSPropertyOperations.js:31
  if (document.documentElement.style.cssFloat === undefined) {
                              ^
TypeError: Cannot read property 'style' of undefined
```

When you consider how we'd have to mock out all the DOM functionality that React uses, the mocks could get quite heavy. In the past I've pulled in `jsdom` in this scenario, to act as a really lightweight browser.

exogen:
To follow up, I think it's totally doable to ditch Karma and run the tests in Node. But I'd rather stick with Karma and use `karma start` (without `singleRun`) and then `karma run` for testing during development – the tests are instantaneous that way. It's not just the browser that Karma provides, but it handles the webpack integration (otherwise you need to run webpack first, send its output to a temp file, then have Mocha run that...), it sets up all the globals needed by Mocha, etc.

[...]

PR #289, opened 2015-07-23
Add test-dev command for much faster test feedback during development

exogen:
Until we decide whether to ditch Karma, which would be a bigger effort, this command can help address #285 in the meantime.


ianobermiller:
Looks good. Can you rebase before merging?

exogen:
@ianobermiller Done! merging now.

PR #294, opened 2015-07-27
Fix errors emitted by latest version of Flow

exogen:
* Add missing return value annotation
* Migrate away from `Object<K, V>` syntax which is no longer supported
* Fix annotation of `existingKeyMap`, whose values are boolean (I think)
* Fix potential null + string error (`keyframesPrefixed`)

Unfortunately there is no way to specify which version of Flow we depend on, since it's not JS and exists outside of npm land.

Done in preparation for #270.

### DIFF
--- a/modules/config.js
+++ b/modules/config.js
@@ -12,7 +12,7 @@ module.exports = {
     return typeof _matchMediaFunction === 'function';
   },
 
-  matchMedia (query: string) {
+  matchMedia (query: string): Object {
     return _matchMediaFunction(query);
   },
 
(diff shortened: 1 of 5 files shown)


exogen:
To get Flow as a dependency and not have to install out-of-band, we could use `flow-bin` on npm. It's unofficial which I don't like, but it's got a shout out on the official repo.

ianobermiller:
Someone created an npm module for flow, flow-bin. We can use that, and add
an npm script "flow" for it, and even add it to non-windows CI. Woot!

exogen:
@ianobermiller Already on it, see #296 :)

exogen:
Closing in favor of #296, which includes this.

PR #296, opened 2015-07-27
Run Flow during lint and CI

### LINKED ISSUE #270: Run flow in the continuous integration tests

exogen:
#294 should be merged first, which fixes errors thrown with the current version of Flow – although, I'm seeing different errors when running Flow this way, that weren't thrown when I submitted #294. Can investigate once that is merged to master and I'll rebase this.

Closes #270


exogen:
Extra errors were due to an accidentally edited line in flow config. #294 now fixes all errors seen in this branch.

modules/config.js
```diff
@@ -12,7 +12,7 @@ module.exports = {
     return typeof _matchMediaFunction === 'function';
   },
 
-  matchMedia (query: string) {
+  matchMedia (query: string): Object {
```
ianobermiller: Should be an object with addListener and removeListener functions: https://developer.mozilla.org/en-US/docs/Web/API/MediaQueryList

The more specific we get, the more bugs flow can catch :yellow_heart:

package.json
```diff
@@ -17,7 +17,8 @@
     "dist": "webpack && webpack --config=webpack.config.minified.js",
     "examples": "webpack-dev-server --config examples/webpack.config.js --no-info --content-base examples/",
     "lib": "npm run babel && rimraf lib/__tests__ lib/__mocks__",
-    "lint": "eslint modules",
+    "lint": "eslint modules && npm run lint-flow",
+    "lint-flow": "node -e \"process.exit(process.platform === 'win32' ? 0 : 1)\" || flow check",
```
ianobermiller: maybe just `flow`? i dont see the value in the `lint-` prefix

exogen:
@ianobermiller sgtm, updated!

modules/config.js
```diff
@@ -12,7 +12,10 @@ module.exports = {
     return typeof _matchMediaFunction === 'function';
   },
 
-  matchMedia (query: string) {
+  matchMedia (query: string): {
+    addListener: () => void,
```
ianobermiller: this should cause flow to fail, but I think it doesn't because flow loses all info on bind. oh well, good enough

package.json
```diff
@@ -16,8 +16,9 @@
     "babel": "rimraf lib && babel --stage 0 modules/ -d lib/",
     "dist": "webpack && webpack --config=webpack.config.minified.js",
     "examples": "webpack-dev-server --config examples/webpack.config.js --no-info --content-base examples/",
+    "flow": "node -e \"process.exit(process.platform === 'win32' ? 0 : 1)\" || flow check",
     "lib": "npm run babel && rimraf lib/__tests__ lib/__mocks__",
-    "lint": "eslint modules",
+    "lint": "eslint modules && npm run flow",
```
ianobermiller: lets leave lint alone, but add `npm run flow` to `prepublish`

exogen: Hmmm, IMO that's asking for people to forget to run it before they commit. CI would catch it down the road, but it would be nice to not have to run yet another command while you're developing. Since eslint (+ other linters) already have a bunch of rules that amount to type checks, they fit in the same category in my mind.

(Not many projects are using `flow-bin`, but at least one of them does have flow in its lint command.)

I can be swayed either way but curious what others think.

ianobermiller: We should have a slightly separate path for dev vs CI then; I don't mind flow being part of lint, but `flow check` is far too slow to be part of lint. `flow status`, which starts the server, is much much faster.

PR #298, opened 2015-07-28
Resolve styles on components found in props and children as a function

### LINKED ISSUE #276: Resolve styles on components passed as props
sometimes a component accepts a nested element as a property (e.g. [material-ui left-nav header property](http://material-ui.com/#/components/left-nav)), further it is allowable to specify the element in-line (vs a reference to a formally defined class).

here is an example (note the syntax in which the `header` property is passed to `MuiLeftNav`):

```
import React, {Component} from 'react'
import MuiLeftNav from 'material-ui/lib/left-nav'
import MenuItem from 'material-ui/lib/menu/menu-item'
import Radium from 'radium'

@Radium
class LeftNav extends Component {

  render() {
    let menuItems = [
      { route: 'home', text: 'Home' },
      { route: 'sign-in', text: 'Sign In' },
      { route: 'sign-up', text: 'Sign Up' },
      { type: MenuItem.Types.SUBHEADER, text: 'Resources' },
      {
         type: MenuItem.Types.LINK,
         payload: 'https://github.com/tony-kerz/react-scratch',
         text: 'GitHub'
      }
    ]

    return(
      <MuiLeftNav
        ref="leftNav"
        docked={false}
        menuItems={menuItems}
        header={ 
          <div
            onClick={this._onNavHeaderClick}
            style={styles.foo}> // <---------------------------------------------------- here...
            foo
          </div>
        }
        onChange={this._onLeftNavChange}
      />
    )
  }

  toggle() {
    this.refs.leftNav.toggle()
  }

  _onLeftNavChange = (e, key, payload) => {
    console.log('left-nav: olnc: e=%o, key=%o, payload=%o', e, key, payload)
    this.context.router.transitionTo(payload.route)
  }

  _onNavHeaderClick = () => {
    console.log('left-nav: ohc')
    this.context.router.transitionTo('home')
    this.refs.leftNav.close()
  }

  static contextTypes = {
    router: React.PropTypes.func
  }
}

let styles = {
  foo: {
    background: 'red',
    color: 'blue',
    ':hover': {
      cursor: 'crosshair',
      background: 'green'
    }
  }
}

export default LeftNav
```

however, radium doesn't seem to be working it's magic for the `style` property passed to that `div` for the `header` property.

i was thinking maybe to use something like `Radium.wrap` around the `div` but looks like `Radium.wrap` is being retired.

any syntax i can use for situations like this or would i need to create a formal class for the `div` and decorated it with `@Radium`
to get this to work...?

### LINKED ISSUE #281: Resolve styles on elements returned from function as children
Unless there's some user error, I think Radium does something that breaks react-motion. React-motion works by using Spring components to specify animation behavior. With the @Radium decorator I get the following Javascript error from the Spring's render method.

```
Uncaught TypeError: this.props.children is not a function
```
in the following code
```
render: function render() {
    var currVals = this.state.currVals;

    return _react2['default'].Children.only(this.props.children(currVals));
  }
```
React-motion works without Radium and Radium works without react-motion. Here's my render method:
```
render() {
    return (
      <Spring className="box" endValue={{val: this.state.isClosed ? 1 : 0, config: [90, 18]}}>
        {({val}) =>
            <img src={this.props.img}
                 style={[styles.img, this.imgStyle(val)]}
                 onMouseDown={this.handleMouseDown.bind(this)}/>
        }
      </Spring>
    );
  }
```
Any thoughts?

ianobermiller:
Also introduces new integration-style tests, that test Radium from the
consumer perspective (which should perhaps be most of the tests….).

Fixes #276.
Fixes #281.

### DIFF
--- a/modules/__tests__/radium-test.js
+++ b/modules/__tests__/radium-test.js
@@ -1,0 +1,183 @@
+var Radium = require('index.js');
+var React = require('react/addons');
+
+var {Component} = React;
+var TestUtils = React.addons.TestUtils;
+
+var getRenderOutput = function (element) {
+  var renderer = TestUtils.createRenderer();
+  renderer.render(element);
+  return renderer.getRenderOutput();
+};
+
+describe('Radium blackbox tests', () => {
+  it('merges styles', () => {
+    @Radium
+    class TestComponent extends Component {
+      render () {
+        return (
+          <div style={[
+            {color: 'blue'},
+            {background: 'red'}
+          ]} />
+        );
+      }
+    }
+
+    var output = getRenderOutput(<TestComponent />);
+
+    expect(output.props.style).to.deep.equal(
+      {color: 'blue', background: 'red'}
+    );
+  });
+
+  it('resolves styles on props', () => {
+    class InnerComponent extends Component {}
+
+    @Radium
+    class TestComponent extends Component {
+      render () {
+        return (
+          <InnerComponent header={
+            <div style={[{color: 'blue'}, {background: 'red'} ]}/>
+          } />
+        );
+      }
+    }
+
+    var output = getRenderOutput(<TestComponent />);
+
+    expect(output.props.header.props.style).to.deep.equal(
+      {color: 'blue', background: 'red'}
+    );
+  });
+
+  it('resolves styles on props', () => {
+    class InnerComponent extends Component {
+      render () {
+        return this.props.stuff;
+      }
+    }
+
+    @Radium
+    class TestComponent extends Component {
+      render () {
+        return (
+          <InnerComponent stuff={
+            <div style={[
+              {color: 'blue'},
+              {background: 'red', ':active': {color: 'green'}}
+            ]} />
+          } />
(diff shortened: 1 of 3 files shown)


ianobermiller:
@alexlande or @ryan-roemer anyone mind taking a look? Just like to have a second set of eyes on this.

modules/__tests__/radium-test.js
```diff
@@ -0,0 +1,158 @@
+var Radium = require('index.js');
+var React = require('react/addons');
+
+var {Component} = React;
+var TestUtils = React.addons.TestUtils;
+
+// NOTE: Many of these tests use mouseDown and active styles. For some reason,
```
alexlande: This, maybe? https://github.com/facebook/react/issues/1297

Current workaround is `TestUtils.SimulateNative.mouseOver`, if so. Will be fixed in 0.14.

ianobermiller: Good to know, thanks!

Issue #132, opened 2015-05-07
`@media print` doesn't work when calling `window.print` in Chrome


bobbyrenwick:
I have made a [jsbin](http://output.jsbin.com/jequceropa/9) to demonstrate the behaviour. When pressing `cmd + p`, both the Radium styled text and the CSS styled text turn red. But when you press the print button, only the CSS styled text turns red.

It seems to be related to the difference between calling `window.print` in JS and pressing `cmd + p` in Chrome.

At the moment I am working around this by having a flux store that listens to `matchMedia` changes, and an action to explicitly set the state.

```js
function print() {
    PrintStateActionCreators.clickedPrint();
    _.defer(() => print());
};
```

```js
window.matchMedia("print").addListener((mql) => {
    this.state = this.state.set("printing", mql.matches);
    this.emitChange();
});
```

The `defer` is the key, because it gives React time to rerender the DOM, given the new styling. This solution is somewhat lacking though because the user will see the print styled component behind the print preview dialog. Any suggestions for how I could avoid this situation?

ianobermiller:
Here is an updated bin that uses font-weight instead since it is easier to see: http://jsbin.com/guzul/3/edit?html,css,js. Its also not anonymous, so it won't time out.

I did some more digging, and this problem is not unique to Radium; it occurs whenever you update the document asynchronously in response to matchMedia for `@media print`. Since React's render calls are always batched, you need the defer as you've found out. The thing is, browsers are not consistent. Using the above jsbin:

Is "Test Radium" bold? (all testing on Mac)

Browser | <kbd>cmd+p</kdb> | PRINT button | PRINT defer button
------|-----|-----|----|-----
**Chrome** | :x: | :x: | :white_check_mark: 
**Firefox** | :x: | :x: | :x: 
**Safari** | :white_check_mark: |:x: (see note) | :x: (see note)

Note: Safari doesn't support `window.print`

Unfortunately, it looks like this is a limitation of inline styles : (. We should add this to the readme.

bobbyrenwick:
Thanks for doing some more digging. That's pretty thorough!

There was one way I was thinking of working around it by making the new `Style` component able to render media queries, rather than just classes. That way we could keep all the logic within the components?

alexlande:
That's really interesting. I can't believe we don't call this out in the docs (and will rectify that soon), but the `Style` component can do media queries: https://github.com/FormidableLabs/radium/blob/master/examples/app.jsx#L84-L116

[...]

PR #299, opened 2015-07-30
Bringing print styles into the Radium ecosystem

bobbyrenwick:
This is an attempt to handle print styles within the Radium ecosystem. This might not align with your vision so before I add tests and learn how to do flow annotations I thought I'd check what you guys think of this.

Basically this arises from the issues described in #132. The end result of that issue is that in order to do print styles properly, you need CSS styles with a media query in. Given that I initially went with creating a `<PrintStyle>` component described in a short [blog post](https://medium.com/@bobbyrenwick/print-styles-with-radium-dc26717df387).

But I just started doing some more serious work on print styles in our application and this pulled up another issue not listed in that blog post in that when you have an item that is repeated, the styles are repeated for each item. This seemed really wrong so I thought of another solution that works similarly to Relay, in that it looks at all the components that are currently rendered and outputs a `<style>` tag with all the print CSS needed for that.

I have no idea how to do that on every React render so I've gone for something similar but without being related to what's currently rendered. Anything that gets enhanced with Radium, and has `printStyles` as a static property, will get given a className on render and you can use a `<PrintStyle>` component that contains all the print styles in a media query once in your app.

What do you guys think @ianobermiller @alexlande? Is this something that you want `radium` to do, or is it too much magic?

P.S Working more on the print styles also revealed that in order to override the inline styles, `!important` should be used on each of the styles. At the moment this'll break when using unitless values but that could be fixed if you guys like this approach


ianobermiller:
I like the idea of solving for print styles, and I like that this implementation works with server-side rendering. I'm not crazy about automatically putting the className on the root item and not being able to style other items.

What about an API more like this?

```as
@Radium
class MyComponent extends Component {
  static printStyles = {
    foo: {width: '100%'},
    bar: {color: 'black'}
  }

  render() {
    return (
      <div className={this.printStyleClass.foo}>
        <div className={this.printStyleClass.bar} />
      </div>
    );
  }
}
```

This is less magical, and thought it leaks the implementation (the classname bit), and pollutes the component scope (the this.printStyleClass bit), it is easy to use and flexible.

bobbyrenwick:
Hey @ianobermiller I'm glad you think that this is within Radium's remit! I think you're suggested solution is really really nice :trophy: I'll implement this today.

I was also thinking it'd be sensible to prefix each class with `Radium-` to avoid collisions.

bobbyrenwick:
@ianobermiller that API is now implemented, with flow annotations and a test to ensure that instances of the component get the correct `printStyleClass` object.

I would also like to add a test that the `PrintStyle` component works but I'll probably wait until #298 is merged because it seems that's the style of test you're heading towards.

Running the tests locally they pass, any idea why the babel loader doesn't seem to be working on karma runs?

bobbyrenwick:
@ianobermiller the tests are fixed by pinning explicitly to the version of `karma-webpack` and `webpack` that were careted in `package.json`. After digging some more I found https://github.com/webpack/karma-webpack/issues/65 and it seems pinning `webpack` was unnecessary so I can change that if you'd like.

Are you OK with the test example not using the `Radium` decorator or would you like to turn on `stage=0` in the tests? I think it's better to leave it as it is because people probably won't want to use `stage=0` in production.

package.json
```diff
@@ -57,7 +57,7 @@
     "react": ">=0.13.3",
     "sinon": "^1.15.3",
     "sinon-chai": "^2.8.0",
-    "webpack": "^1.9.11",
+    "webpack": "1.9.11",
```
ianobermiller: yeah, please unpin webpack.

package.json
```diff
@@ -48,7 +48,7 @@
     "karma-phantomjs-launcher": "^0.2.0",
     "karma-phantomjs-shim": "^1.0.0",
     "karma-sinon-chai": "^1.0.0",
-    "karma-webpack": "^1.5.1",
+    "karma-webpack": "1.5.1",
```
ianobermiller: sounds like updating karma and karma-webpack would fix it too? https://github.com/webpack/karma-webpack/issues/65#issuecomment-126675635

modules/components/print-style.js
```diff
@@ -0,0 +1,46 @@
...
+    return {
+      styles: printStyles.getPrintStyles()
+    };
+  },
+
+  render: function () {
+    return (
+      <Style
+        rules={
+          {
```
ianobermiller: nit: put on same line as `rules={`, same for closing

modules/__tests__/radium-test.js
```diff
@@ -180,4 +181,45 @@ describe('Radium blackbox tests', () => {
...
+    };
+    Radium(TestComponent2);
+
+    var output = TestUtils.renderIntoDocument(<PrintStyle />);
+
+    var style = React.findDOMNode(
+      TestUtils.findRenderedDOMComponentWithTag(output, 'style')
+    );
+
+    expect(style.innerText).to.deep.equal(
```
ianobermiller: no need for deep.equal on a string

docs/api/README.md
```diff
@@ -223,3 +224,55 @@ A string that any included selectors in `rules` will be appended to. Use to scop
   />
 </div>
 ```
+
+## Print Style
+
+In order to fully support all print styling eventually, it is necessary to use CSS because of browser differences described in #132. Radium allows you to do this easily by specifying print styles as static properties of your components.
```
ianobermiller: "In order to fully support print styling it is necessary to use CSS because of browser differences, as described in #132. Radium allows you to do this easily by specifying print styles as static properties of your components.

docs/api/README.md
```diff
@@ -223,3 +224,55 @@ A string that any included selectors in `rules` will be appended to. Use to scop
...
+
+With ES7 decorators and static class properties:
+
+```as
+@Radium
+class MyComponent extends React.Component {
+  static displayName = 'MyComponent';
```
ianobermiller: No need for this; see other comment

modules/print-styles.js
```diff
@@ -0,0 +1,70 @@
+/* @flow */
+
+var allPrintStyles = {};
+var listeners = [];
+
+var addListener = function (listener: Function): void {
```
ianobermiller: since we now implement this twice in two different modules, should we move this to a helper? (also in mouse-up-listener)

modules/print-styles.js
```diff
@@ -0,0 +1,70 @@
+/* @flow */
```
ianobermiller: all indentation in this file is off, should be two spaces

modules/print-styles.js
```diff
@@ -0,0 +1,70 @@
...
+    if (listenerIndex > -1) {
+        listeners.splice(listenerIndex, 1);
+    }
+};
+
+var emitChange = function () {
+    listeners.forEach(listener => listener());
+};
+
+var importantValues = function (styleObj) {
```
ianobermiller: rename: _appendImportant, or _appentImportantToEachValue

Also put `_` at the beginning of every non-exported function

modules/print-styles.js
```diff
@@ -0,0 +1,70 @@
...
+    var printStyleClass = {};
+
+    Object.keys(Component.printStyles).forEach((key) => {
+        var styles = Component.printStyles[key];
+        var className = `Radium-${Component.displayName}-${key}`;
+        allPrintStyles[`.${className}`] = importantValues(styles);
+        printStyleClass[key] = className;
+    });
+
+    // Allows for lazy loading of JS that then calls Radium to update the
```
ianobermiller: nice

modules/components/print-style.js
```diff
@@ -0,0 +1,46 @@
+var React = require('react');
+
+var Style = require('./style.js');
+var printStyles = require('../print-styles.js');
+
+var PrintStyle = React.createClass({
+
```
ianobermiller: no extra line

modules/components/print-style.js
```diff
@@ -0,0 +1,46 @@
...
+  componentDidMount: function () {
+    printStyles.addListener(this.onChange);
+  },
+
+  componentWillUnmount: function () {
+    printStyles.removeListener(this.onChange);
+  },
+
+  onChange: function () {
```
ianobermiller: prefix onChange and getStylesState with _

modules/components/print-style.js
```diff
@@ -0,0 +1,46 @@
+var React = require('react');
+
+var Style = require('./style.js');
+var printStyles = require('../print-styles.js');
+
+var PrintStyle = React.createClass({
```
ianobermiller: would it be clearer to call this PrintStyleSheet?

ianobermiller:
Nice work so far, thanks for making the API I came up with off the top of my head come to life :)

package.json
bobbyrenwick: Yea you're right, I thought that a fresh `npm install` would be installing karma 0.13.3 but I forgot that's not how carets work for 0.x.x is it!

package.json
bobbyrenwick: :+1:

docs/api/README.md
bobbyrenwick: Wow I've just reread what I wrote. :+1: on the correction!

docs/api/README.md
bobbyrenwick: :+1: I'm still thinking in ES5!

modules/print-styles.js
bobbyrenwick: Ah yea! Any reason there's no `indent: [2, 2]` in the `.eslintrc`?

modules/components/print-style.js
bobbyrenwick: I was keeping in line the with `Style` component but I agree that `PrintStyleSheet` actually encapsulates the purpose of this component better than `PrintStyle`.

modules/print-styles.js
bobbyrenwick: In order to reuse the `subscribe` implementation we'd have to be able to pass extra functions in to allow for [adding](https://github.com/FormidableLabs/radium/blob/master/modules/mouse-up-listener.js#L17-L20) and [removing](https://github.com/FormidableLabs/radium/blob/master/modules/mouse-up-listener.js#L27-L30) the mouse up listener given certain information.

I think that'll end up being more complicated than the current, slightly duplicated approach.

bobbyrenwick:
Hey @ianobermiller, no worries about making the API come to life! I'm in need of it so I might as well make it happen! I've done most of the things you've suggested and left a couple of comments that need your input.

docs/api/README.md
```diff
@@ -223,3 +224,54 @@ A string that any included selectors in `rules` will be appended to. Use to scop
...
+## PrintStyleSheet component
+
+In order to fully support print styling it is necessary to use CSS because of browser differences, as described in #132. Radium allows you to do this easily by specifying print styles as static properties of your components.
+
+With ES7 decorators and static class properties:
+
+```as
+@Radium
+class MyComponent extends React.Component {
+
```
ianobermiller: remove empty line

docs/api/README.md
```diff
@@ -223,3 +224,54 @@ A string that any included selectors in `rules` will be appended to. Use to scop
...
+    return (
+      <div className={this.printStyleClass.wrapper}>
+        <p className={this.printStyleClass.text}>I'm red on print</p>
+      </div>
+    );
+  }
+}));
+```
+
+Somewhere in near the root of your app, create a `<PrintStyleSheet />` component and it will render a style tag containing all the CSS needed for printing, wrapped in a `@media print` query.
```
ianobermiller: Add a note, something like "You should only render `<PrintStyleSheet />` once. It will contain all the print styles for every component."

ianobermiller: Also, you should probably `require` PrintStyleSheet last, right? Otherwise you won't capture all the print styles on the first render pass, which is necessary with SSR (server side rendering).

modules/__tests__/radium-test.js
```diff
@@ -180,4 +181,45 @@ describe('Radium blackbox tests', () => {
...
+    Radium(TestComponent2);
+
+    var output = TestUtils.renderIntoDocument(<PrintStyleSheet />);
+
+    var style = React.findDOMNode(
+      TestUtils.findRenderedDOMComponentWithTag(output, 'style')
+    );
+
+    expect(style.innerText).to.equal(
+      '@media print{.Radium-TestComponent-foo{color: blue !important;}.Radium-TestComponent-bar{background: red !important;}.Radium-TestComponent2-main{color: black !important;}}' // eslint-disable-line max-len
```
ianobermiller: instead of disabling the lint, just break it up into multiple concatenated lines

modules/print-styles.js
ianobermiller: Fair enough. Though I still think the subscribe method returning an object with remove is a better API than addListener/removeListener

modules/print-styles.js
ianobermiller: Oops, that should definitely be in there. Feel free to add it, or we'll get to it eventually :)

ianobermiller:
One more thing, just to definitively answer the question of bloat; can you measure the gzipped distribution size before and after your changes?

There are some unreleased changes already committed, so to test the before, update to master, and run:

```
npm run dist
cat dist/radium.min.js | gzip | wc -c
```

Which for me, on master, is 4,768 bytes. For comparison, the previous release (only a single feature commit back) was 4,714 bytes.

modules/print-styles.js
bobbyrenwick: :+1: added

modules/print-styles.js
bobbyrenwick: OK I'll change it to be done like that.

docs/api/README.md
bobbyrenwick: Note added.

As for making the `require` of `PrintStyleSheet` last, I'm not sure about that, because the key is  server side rendering is like, but the key is when `getInitialState` is called for `PrintStyleSheet`. Doesn't that happen when at the `React.createElement(PrintStyleSheet, ...)` call? So I think it should be OK, but I could be wrong.

docs/api/README.md
bobbyrenwick: :+1:

modules/__tests__/radium-test.js
bobbyrenwick: :+1: it's way better like that

modules/print-styles.js
bobbyrenwick: Done!

bobbyrenwick:
```
$ cat dist/radium.min.js | gzip | wc -c
5099
```

bobbyrenwick:
So a 7% increase in size. Is that OK?

ianobermiller:
Yes of course it is ok, I was mearly curious. 331 bytes for proper print style support is pretty awesome!

Looks really great, thanks for going through so many iterations, I am a bit of a sticker on code reviews. Worth it to keep the code something to be proud of!

bobbyrenwick:
Brilliant, I'm glad to have helped.

No worries about the code review! I'm exactly the same when it comes to our code base. Sometimes it's just difficult to pick up on coding conventions, but the more that eslint can help the better.

PR #305, opened 2015-08-04
Un-prefix values before checking isUnitlessNumber

### LINKED ISSUE #297: Prefixer should unprefix values before checking isUnitlessNumber
Otherwise, if you explicitly pass WebkitFlex: 1, it will end up as WebkitFlex: '1px'. The current behavior, combined with double resolving, is causing the bug #293.

AnSavvides:
Added a test to demonstrate this issue to begin with & then made the necessary changes to make it work. If anyone can think of a more elegant solution, feel free to say so.

Closes #297.

### DIFF
--- a/modules/prefixer.js
+++ b/modules/prefixer.js
@@ -251,7 +251,7 @@ var getPrefixedPropertyName = function (property) {
 var _getUnprefixedProperty = function (property) {
   var noPrefixProperty = property;
 
-  vendorPrefixes.forEach(prefix => {
+  vendorPrefixes.some(prefix => {
     // Let's check if the property starts with a vendor prefix
     if (property.indexOf(prefix) === 0) {
       noPrefixProperty = noPrefixProperty.replace(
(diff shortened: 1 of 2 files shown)


modules/prefixer.js
```diff
@@ -242,15 +244,44 @@ var getPrefixedPropertyName = function (property) {
...
 };
 
+// We are un-prefixing values before checking for isUnitlessNumber,
+// otherwise we are at risk of being in a situation where someone
+// explicitly passes something like `MozBoxFlex: 1` and that will
+// in turn get transformed into `MozBoxFlex: 1px`.
+var _getUnprefixedProperty = function (property) {
+  var noPrefixProperty = property;
+
+  vendorPrefixes.forEach(prefix => {
```
ianobermiller: do `some` instead of `forEach` and return true once one of the prefixes matches to shortcircuit

ianobermiller:
Looks great!

modules/prefixer.js
AnSavvides: Ah yes, I didn't even think of that - I was too caught up about the fact that you can't use a `break` with `forEach`! Fix incoming :)

PR #306, opened 2015-08-04
Prevents errors from getters that do not have setters.

rolandpoulter:
I'm using radium with material-ui which seems to be defining a getter without a setter. It causes this code to break. This is a much safer way to copy properties.


ianobermiller:
Nice, thanks for the PR!

PR #310, opened 2015-08-11
Adding enhancer tests for better coverage

### LINKED ISSUE #264: Add tests for enhancer static prop transferring
We are missing test coverage for some fixes landed in #241 and #255.

See missing coverage: https://coveralls.io/builds/3057610/source?filename=modules%2Fenhancer.js#L60

AnSavvides:
This closes #264 by bringing in better test coverage to the enhancer.

### DIFF
--- a/modules/__tests__/enhancer-test.js
+++ b/modules/__tests__/enhancer-test.js
@@ -165,4 +165,28 @@ describe('Enhancer', () => {
     expect(Enhanced.defaultProps).to.deep.equal({ foo: 'bar' });
   });
 
+  it('copies methods across to top level prototype', () => {
+    var Composed = React.createClass({
+
+      getStyles: function () {
+        return [{ color: 'black' }];
+      },
+
+      render: function () {
+        return (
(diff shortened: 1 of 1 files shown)


modules/__tests__/enhancer-test.js
```diff
@@ -3,7 +3,8 @@ var Enhancer = require('inject!enhancer.js')({
   './resolve-styles.js': resolveStyles
 });
 
-var {Component} = require('react');
+var React = require('react/addons');
```
ianobermiller: we should probably be consistent and always require react/addons in tests, might help a bit with speed (just conjecture).

ianobermiller:
Sweet! :rocket:

modules/__tests__/enhancer-test.js
AnSavvides: After some crude testing it doesn't look like it makes much of a difference (if it does, it's very marginal), but we should do this for consistency's sake anyway - so here it is: https://github.com/FormidableLabs/radium/pull/312

PR #312, opened 2015-08-12
Consistently require react/addons

AnSavvides:
This could have a slight speed-up effect, but doing some crude testing hasn't shown this to be of a real significance. In any case, it's good to be consistent in how we do things!

### DIFF
--- a/modules/__tests__/keyframes-test.js
+++ b/modules/__tests__/keyframes-test.js
@@ -1,8 +1,7 @@
 var styleElement;
 var exenv;
 
-var React = require('react/addons');
-var {Component} = React;
+var {Component} = require('react/addons');
 
 class MyComponent extends Component { }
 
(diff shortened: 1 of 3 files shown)


modules/__tests__/keyframes-test.js
```diff
@@ -1,7 +1,8 @@
 var styleElement;
 var exenv;
 
-var {Component} = require('react');
+var React = require('react/addons');
```
ianobermiller: You can keep these as `var {Component} = require('react/addons');` any good reason to change it to two lines?

AnSavvides: Ha, that's a great point actually... Got carried away from `resolve-styles` where we actually make use of `React`! Fix incoming any second now :)

ianobermiller:
:heart:

PR #316, opened 2015-08-14
Copy all static props

### LINKED ISSUE #313: Static class methods disappear in IE10 when using Radium
Static class methods are gone after wrapping the component class in `Radium(Foo)` in IE10 (using Babel to convert ES6):

*Simple example:*
```javascript
import Radium from 'radium';
import React from 'react/addons';

const TYPE_FOO = 'foo';

class Foo extends React.Component {
    static getTypeFoo() {
        return TYPE_FOO;
    }

    render() {
        return <div>Foo</div>
    }
}

export default Radium(Foo);
```

*Result in IE10:*
```javascript
Foo.getTypeFoo() // IE10 throws: Object doesn't support property or method 'getTypeFoo'
```

The above works fine in Chrome. It also works in IE10 when removing `Radium()` decorator.

AnSavvides:
Fixes bug with only copying a certain set of static properties - now all properties are copied.

Closes #313

### DIFF
--- a/modules/__tests__/enhancer-test.js
+++ b/modules/__tests__/enhancer-test.js
@@ -155,14 +155,16 @@ describe('Enhancer', () => {
     });
   });
 
-  it('manually populates static properties for IE <10', () => {
+  it('manually populates all static properties for IE <10', () => {
     class Composed extends Component { }
 
     Composed.defaultProps = { foo: 'bar' };
+    Composed.someOtherProp = { bar: 'foo' };
 
     var Enhanced = Enhancer(Composed);
 
     expect(Enhanced.defaultProps).to.deep.equal({ foo: 'bar' });
+    expect(Enhanced.someOtherProp).to.deep.equal({ bar: 'foo' });
   });
 
   it('copies methods across to top level prototype', () => {
--- a/modules/__tests__/enhancer-test.js
+++ b/modules/__tests__/enhancer-test.js
@@ -157,7 +157,7 @@ describe('Enhancer', () => {
 
   it('manually populates all static properties for IE <10', () => {
     class Composed extends Component {
-      static staticMethod() {
+      static staticMethod () {
         return { bar: 'foo' };
       }
     }
--- a/modules/__tests__/enhancer-test.js
+++ b/modules/__tests__/enhancer-test.js
@@ -156,15 +156,18 @@ describe('Enhancer', () => {
   });
 
   it('manually populates all static properties for IE <10', () => {
-    class Composed extends Component { }
+    class Composed extends Component {
+      static staticMethod() {
+        return { bar: 'foo' };
+      }
+    }
 
     Composed.defaultProps = { foo: 'bar' };
-    Composed.someOtherProp = { bar: 'foo' };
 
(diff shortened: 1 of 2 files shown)


modules/enhancer.js
```diff
@@ -48,30 +48,15 @@ var enhanceWithRadium = function (ComposedComponent: constructor): constructor {
...
-    'contextTypes',
-    'childContextTypes'
-  ];
-
-  staticKeys.forEach((key) => {
-    if (ComposedComponent.hasOwnProperty(key)) {
-      RadiumEnhancer[key] = ComposedComponent[key];
+  // This also fixes React Hot Loader by exposing the original components top level
+  // prototype methods on the Radium enhanced prototype as discussed in #219.
+  Object.keys(ComposedComponent.prototype).forEach(key => {
```
Zenwolf: ES6 class properties are not enumerable with `Object.keys`, as per the spec. They are defined with `Object.defineProperty` where `enumerable` is false by default. So the keys will not show up using `Object.keys()` and will not be copied. When you assign a vanilla property with `MyComponent.foo = someVal`, it will show up because the old default for `enumerable` when not using `Object.defineProperty` is true.

ianobermiller: `Object.getOwnPropertyNames` should do the trick, eh?

AnSavvides: Thanks for pointing that out @Zenwolf - I completely missed that! `Object.getOwnPropertyNames` is spot on @ianobermiller! :)

modules/__tests__/enhancer-test.js
```diff
@@ -155,14 +155,19 @@ describe('Enhancer', () => {
     });
   });
 
-  it('manually populates static properties for IE <10', () => {
-    class Composed extends Component { }
+  it('manually populates all static properties for IE <10', () => {
+    class Composed extends Component {
+      static staticMethod() {
```
ianobermiller: need space after `staticMethod`, lint is failing

AnSavvides: Hah, I literally just pushed a fix for this as you typed this - I only just realized this was failing. It'd be nice if a fail triggered an e-mail alert of some sort :)

Yeah I am not a fan of this lint rule - `fn()` > `fn ()` as far as I'm concerned!

ianobermiller:
nice, thanks. Yes, maybe github or travis has a configuration for sending emails on failed PR checks.

modules/__tests__/enhancer-test.js
```diff
@@ -155,14 +155,19 @@ describe('Enhancer', () => {
...
-  it('manually populates static properties for IE <10', () => {
-    class Composed extends Component { }
+  it('manually populates all static properties for IE <10', () => {
+    class Composed extends Component {
+      static staticMethod () {
+        return { bar: 'foo' };
+      }
+    }
 
     Composed.defaultProps = { foo: 'bar' };
```
ianobermiller: Since we're using babel anyway, move defaultProps to the class:

```as
class Composed extends Component {
  static staticMethod () {
    return { bar: 'foo' };
  }

  defaultProps = {foo: 'bar'}
}

PR #321, opened 2015-08-18
getStateKey should return "main" when empty string is passed in

hledley:
Previously an empty string key would result in 'main'.  Functionality changed in this commit: https://github.com/FormidableLabs/radium/pull/287

### DIFF
--- a/modules/get-state-key.js
+++ b/modules/get-state-key.js
@@ -1,7 +1,7 @@
 /* @flow */
 
 var getStateKey = function (elementKey: ?string): string {
-  return elementKey === null || elementKey === undefined ?
+  return elementKey === null || elementKey === undefined || elementKey === '' ?
     'main' :
     elementKey.toString();
 };


AnSavvides:
I will let @ianobermiller chime in whether that was intentional or not given that he made that change, but including a test case to cover this scenario alongside your changes would be great.

ianobermiller:
This wasn't intential, but what is wrong with a key that is the empty string? I guess it is harder to debug, but technically speaking it works fine (even React doesn't mind `key=""`). 
I don't feel strongly either way. Perhaps add a test that also uses `getState`.

https://jsbin.com/veqoba/edit
```as
var Example = React.createClass({
  render() {
    return (
      <div>{[
        <div key="">wow</div>,
        <div key="fo">bam</div>
      ]}</div>
    );
  }
});

React.render(<Example />, document.getElementById('main'));

hledley:
Ah, I hadn't thought about having a key actually be "".

My initially reason for this pull request is I had been using "", and upon updating to the most recent version of radium I noticed all my calls to radium.getState were broken. I've since switched to using undefined, so this fix is not longer necessary for me but I thought other people may get tripped up by this change.

ianobermiller:
Actually, your old calls to `getState` were rather unintuitive: you supplied key `''`, but had to use `'main'` to check the state. I see why this would seem like a bug. One other odd thing is that if you had both empty string and main you would see an duplicate key error after this change.

ianobermiller:
Thanks, that was the right call, I should have closed this earlier!

PR #327, opened 2015-08-20
:memo: update readme so that links work from npm

basarat:
Currently links are broken at https://www.npmjs.com/package/radium :rose:


ianobermiller:
Only issue with this is that the readme is then tied to `master`, so if you have a change in another branch you can't use the links. I guess I'd rather have NPM work correctly, though. Thanks!

PR #315, opened 2015-08-13
 adding isomorphic example #143

jurgob:

PR #329, opened 2015-08-20
Run examples on universal setup

moret:
This PR allows running the examples on a hapi server to make it easier to detect when React warns about not being able to reuse markup.

It also fixes the first warning found, related to using div's inside p's. The JSX was:
````
<p><div>...</div></p>
````
Resulting in the browser's DOM:
````
<p></p><div>...</div><p></p>
````
I changed it to avoid differences between the server rendering and the browser DOM.

The remaining React warning is related to #201 - Prefixing breaks server rendering.

### DIFF
--- a/examples/app.jsx
+++ b/examples/app.jsx
@@ -94,36 +94,29 @@ var App = React.createClass({
 
     return (
       <div>
-        <p><HoverMessage /></p>
+        <p /><HoverMessage />
 
-        <p><TwoSquares /></p>
+        <p /><TwoSquares />
 
-        <p><Spinner /></p>
+        <p /><Spinner />
 
-        <p>
-          <Button onClick={this._remount}>Unmount and remount</Button>
-        </p>
+        <p /><Button onClick={this._remount}>Unmount and remount</Button>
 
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

File: examples/app.jsx [hunk 1]
```diff
@@ -94,36 +94,29 @@ var App = React.createClass({
 
     return (
       <div>
-        <p><HoverMessage /></p>
```
ianobermiller: Why change all these? `<p>` as a container tag is much nicer.

ianobermiller:
Can you sync with @jurgob, who submitted something very similar in #315? We'll only be able to merge one of them.

examples/app.jsx
ianobermiller: Sorry, I just read the PR description. You're right, and that is crazy. I guess this is why we use `div`s for everything.
image

moret:
Yeah, I liked the container `<p>` too. 8(

I didn't notice there was another PR on the same subject. I will check it out with @jurgob .

ianobermiller:
thanks @moret!

PR #330, opened 2015-08-20
Remove redundant “false”.

iclanzan:


ianobermiller:
haha good catch. thanks!

Issue #334, opened 2015-08-26
Allow style to be a deeply nested array


richardfickling:
I'm just getting started with React and Radium, so it's entirely probable that this suggestion is unnecessary because I'm doing something fundamentally wrong. I'd like for Radium to be able to merge deeply nested style arrays to deal with component inheritance. For example, there's a table in my app:
```
Row = Radium(React.createClass({
  style: {
    main: {
      width: '100%'
    }
  }

  render: ->
    <div style={ [@style.main, @props.style] }> { @props.children } </div>
}))

CustomRow = Radium(React.createClass({
  style: {
    main: {
      height: '86px'
      borderBottom: '1px solid blue'
    }
  }

  render: ->
    <Row style={ [@styles.main, @props.style] }> { @props.children } </Row>
}))

CustomTable = React.createClass({
    render: ->
      <div>
        <CustomRow style={ background: 'red' } />
        <CustomRow style={ background: 'white' } />
      </div>
})
```

This doesn't work because @props.style in the `<Row>`s is an array, so the `<div>`'s style prop is a nested array, which Radium doesn't support (https://github.com/FormidableLabs/radium/blob/master/modules/resolve-styles.js#L49)

Is there a "suggested" way to handle component inheritance in situations like this?

If the solution is adding support for nested arrays, I'm happy to make the PR, but I didn't want to do that without asking for a better solution first.

ianobermiller:
Hmm, yes this does seem like a legitimate use case, and could rather easily be added to mergeStyles if you want to do so.

ianobermiller:
I also have plugin support mostly complete, and this would be a good candidate for a plugin.

almost:
Would there be any downside to including this as a default behaviour? It's very useful when making components and I'd guess it wouldn't add much in the way of complexity of slowdown?

ianobermiller:
Not at all, I'd accept a tested PR with the change.

bobbyrenwick:
:+1: It's great to see this being supported! At the moment we `flatten` out any arrays in our bottom level components in order to achieve the same thing.

PR #344, opened 2015-09-06
Support nested arrays in styles

almost:
As request in #334. I know you were talking about making this a plugin but it seems to be that there wouldn't be any downside to supporting it by default. What do you think?

### DIFF
--- a/modules/__tests__/radium-test.js
+++ b/modules/__tests__/radium-test.js
@@ -32,6 +32,28 @@ describe('Radium blackbox tests', () => {
     );
   });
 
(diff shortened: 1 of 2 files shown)


ianobermiller:
Please add tests.

almost:
Sure, I've just added a test. Let me know if you need anything else

richardfickling:
thanks for making this @almost! Looks like the tests will pass CI if you remove the console.log statement

almost:
Sorry! I really thought I'd removed that. Have taken it out now

PR #331, opened 2015-08-20
Move config to decorator argument

ianobermiller:
First step in a stateless, configurable Radium. This is also going to be the basis for using plugins.

PR #348, opened 2015-09-10
Plugins

ianobermiller:
Refactor Radium to be plugin-based instead of one giant block of code in resolveStyles. This has a nice side-effect of the code being more modular and easier to understand.

This is only a slightly breaking change. Normal usage of Radium works exactly as before, `@Radium`. The main difference is that all config has moved from the stateful `Radium.Config` to a config object passed to `@Radium`, like this: `@Radium({matchMedia: ..., plugins: ...})` (see #331 for moving `matchMedia`). Of course since it is is breaking, it will have to be released as 0.14.0.

Since you don't want to pass the config object every time you use Radium, I suggest creating a small wrapper module, say `ConfiguredRadium`:

```as
var customPlugin = require('./customPlugin');

var _matchMedia = null;

function ConfiguredRadium(component) {
  return Radium({
    matchMedia: _matchMedia,
    plugins: [
      Radium.Plugins.mergeStyleArray,
      Radium.Plugins.checkProps,
      Radium.Plugins.resolveMediaQueries,
      Radium.Plugins.resolveInteractionStyles,
      customPlugin,
      Radium.Plugins.prefix,
      Radium.Plugins.checkProps,
    ]
  })(component);
}

ConfiguredRadium.setMatchMedia = function (matchMedia) {
  _matchMedia = matchMedia;
};

module.exports = ConfiguredRadium;
```

You'll notice something pretty ugly there, which is how to insert in a custom plugin. You have to repeat all the internal plugins. This means if we do something like splitting an internal plugin in half, or some other non-breaking change that changes the list of plugins, it will still break people who are adding their own plugins. I haven't come up with a good solution here yet. We need to allow the following, in order of importance:

1. If the default plugin list changes, and you haven't explicitly opted to replace all the plugins, you should get the updated list automatically (this one fails with the current proposal)
2. Replacing an existing plugin, e.g. replacing `prefix` with `myCustomPrefixer`, or adding a more restrictive/loose `checkProps`
3. Adding a new plugin in the normal pipeline position, before `prefix`
4. Replace all the plugins - Radium can effectively be used as a mechanism for iterating over rendered elements

Any suggestions or comments on the proposal in general?


alexlande:
First impressions are that this looks tremendous. Nice work @ianobermiller :)

ianobermiller:
Yeah sorry, this is a gigantic PR. I wanted to do the dev on a separate branch to avoid having master being super out of date with NPM, and I'm not good enough with GitHub's PR system to do this properly :crying_cat_face:

alexlande:
Ha, I meant tremendous as in "very good or excellent", not "very large or great", but I will admit that it's very large also.

ianobermiller:
After further thought, I think that the current solution is the simplest and the most powerful, and shouldn't have any real drawbacks. A core plugin being added won't get picked up, but that is probably a good thing, as it won't break you if you are using custom plugins. A core plugin being split into two is easy; just keep the existing name, and have it call both. Overall, I'm pretty happy with this approach, and will proceed unless anyone has objections.

alexlande:
:+1:

PR #349, opened 2015-09-10
Fixing bug where no static properties were copied over in IE < 10

bobbyrenwick:
This one again. It seems the implementation in #316 actually stops any static properties being copied over in IE < 10 (@AnSavvides). This had the unfortunate side effect of breaking our app in IE for a few days.

The reasons the tests don't catch it are because we only run the tests in PhantomJS so the bugs that exist in IE < 10 don't materialise. I've changed AppVeyor to use [karma-ie-launcher](https://github.com/karma-runner/karma-ie-launcher) and made IE11 run in IE9 compatibility mode in order to catch the bug. You can see it [catches the bug](https://ci.appveyor.com/project/bobbyrenwick/radium/build/1.0.4/job/x3j8dtfiqr789xfv) on a test run of a branch without this fix in.

### DIFF
--- a/modules/enhancer.js
+++ b/modules/enhancer.js
@@ -56,8 +56,7 @@ var enhanceWithRadium = function (ComposedComponent: constructor): constructor {
 
   // Class inheritance uses Object.create and because of __proto__ issues
   // with IE <10 any static properties of the superclass aren't inherited and
-  // so need to be manually populated. We only want to go over that object's
-  // non-enumerable properties, so use Object.keys.
+  // so need to be manually populated.
   // See http://babeljs.io/docs/advanced/caveats/#classes-10-and-below-
   copyProperties(ComposedComponent, RadiumEnhancer);
 
--- a/modules/enhancer.js
+++ b/modules/enhancer.js
@@ -3,6 +3,16 @@
 var resolveStyles = require('./resolve-styles.js');
 var printStyles = require('./print-styles.js');
 
+var copyProperties = function (source, target) {
+  Object.getOwnPropertyNames(source).forEach(key => {
+    const ignoreKeys = ['type', 'arguments', 'callee', 'caller', 'length', 'name', 'prototype'];
+    if (ignoreKeys.indexOf(key) < 0 && !target.hasOwnProperty(key)) {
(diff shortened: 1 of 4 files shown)


modules/enhancer.js
```diff
@@ -46,8 +46,16 @@ var enhanceWithRadium = function (ComposedComponent: constructor): constructor {
 
   // Class inheritance uses Object.create and because of __proto__ issues
   // with IE <10 any static properties of the superclass aren't inherited and
-  // so need to be manually populated
+  // so need to be manually populated. We only want to go over that object's
+  // non-enumerable properties, so use Object.keys.
   // See http://babeljs.io/docs/advanced/caveats/#classes-10-and-below-
+  Object.keys(ComposedComponent).forEach(key => {
```
ianobermiller: can you abstract this loop into a function?

```
copyProperties(ComposedComponent, RadiumEnhancer);
copyProperties(ComposedComponent.prototype, RadiumEnhancer.prototype);

Comment (bobbyrenwick) [inline modules/enhancer.js]: :+1: done and it fixes the test that was then failing because we weren't copying across static methods defined using `static` keyword within class definitions.

**Comment by bobbyrenwick** [comment]
The tests passed on [AppVeyor](https://ci.appveyor.com/project/bobbyrenwick/radium/build/1.0.8) for my branch and they seem to have just timed out speaking to npm on this run but I don't think I can re-build this PR on the FormidableLabs AppVeyor.

File: modules/enhancer.js [hunk 2]
```diff
@@ -3,6 +3,16 @@
 var resolveStyles = require('./resolve-styles.js');
 var printStyles = require('./print-styles.js');
 
+var copyProperties = function (source, target) {
+  Object.getOwnPropertyNames(source).forEach(key => {
+    var ignoreKeys = ['type', 'arguments', 'callee', 'caller', 'length', 'name', 'prototype'];
```
ianobermiller: Cool, this seems like a good idea ;)

ianobermiller:
:+1:

ianobermiller:
Yeah AppVeyor and NPM is pretty flakey :(

AnSavvides:
:+1:

PR #355, opened 2015-09-15
update jss info

kof:


kof:
Also would be nice to have this comparison in a separate repository

kof:
And maybe some better format because editing this md file is not fun

ianobermiller:
Agreed on both counts! Thanks for updating. Suggestions on a better format? I'd say a repo with one file per library, with some format conducive to multi-line markup.

kof:
not sure what to suggest if simple md will be used ... might be better to generate html  using json configs.

ianobermiller:
Yeah that works for me.

PR #356, opened 2015-09-16
Fix object-assign dependency location

richardfickling:
It's required in modules/enhancer.js


ianobermiller:
Shoot, I should have caught that. Actually, I'll switch it to be spread instead.

PR #360, opened 2015-09-18
Removing dead code

AnSavvides:
As far as I can tell - correct me if I am wrong - this was just dead code.

To begin with, `animationStyleSheet` is set to be `null` and this changes when `isAnimationSupported` is `true` -- within the `keyframes` function, we do the following check:

```js
if (!isAnimationSupported) {
  return name;
}
```

...which happens before the (now removed) `throw`. Therefore, `animationStyleSheet` could never not be `truthy` at this point if `isAnimationSupported` is `true`. If `isAnimationSupported` were `true` and `animationStyleSheet` was to be `null` or a `false`-y value (for whatever reason), it would break here anyway, before ever getting to the `throw` portion of the code:

```js
animationStyleSheet.textContent = '@keyframes {}';
```

We could add a check before the above and if `animationStyleSheet` is `false`-y we could `throw` an `Error`, otherwise proceed?

### DIFF
--- a/modules/keyframes.js
+++ b/modules/keyframes.js
@@ -29,6 +29,12 @@ var keyframes = function (
   keyframeRules: {[percentage: string]: {[key: string]: string|number}},
   componentName?: string
 ): string {
+  // for flow
+  /* istanbul ignore next */
+  if (!animationStyleSheet) {
+    throw new Error('keyframes not initialized properly');
+  }
+
   var name = 'Animation' + animationIndex;
   animationIndex += 1;
 
--- a/modules/keyframes.js
+++ b/modules/keyframes.js
@@ -45,12 +45,6 @@ var keyframes = function (
     }).join('\n') +
     '\n}\n';
 
-  // for flow
-  /* istanbul ignore next */
-  if (!animationStyleSheet) {
-    throw new Error('keyframes not initialized properly');
-  }
-
   animationStyleSheet.sheet.insertRule(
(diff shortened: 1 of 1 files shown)


ianobermiller:
Sorry I couldn't follow your description above. You'll notice the code has a comment: "for flow". This code was added just so flow would pass, which is why your pr is failing. If you want to refactor it in such a way that we don't need the code but flow is also happy, feel free, otherwise you can close this PR.

AnSavvides:
I should've probably explained the above a little better... :) I went for detail but that just made it a tad more confusing!

*tl;dr* - I think the check should be happening earlier. I will make the change, check that `flow` is still happy and push it.

ianobermiller:
:+1:

ianobermiller:
I don't think there is any point to just moving the code. I wouldn't mind putting in some slightly misleading annotations on `animationStyleSheet` to make flow think it always has a value, maybe just `var animationStyleSheet: any = null;` to make flow ignore it?

AnSavvides:
I see what you mean.

Moving it up means it fails *earlier*, if it is to fail, without doing the whole `var keys = ...` bit which is pointless if we are just going to `throw` straight after. On the other hand though, if we are going to `throw`, does it really matter optimizing for that `throw` to happen as soon as possible and minimizing computations?

I think maybe it's better to leave it as is, it's not a big deal.

PR #368, opened 2015-09-25
When hovering, remove :hover from the previous hovered component

### LINKED ISSUE #367: :hover gets wrong events
I have several rows which have hover styles. If I move my mouse over them slowly everything is fine.

When I move the mouse over them quickly I found that the hover styles were remaining. 

I logged the `e.target` of the events that were happening `onMouseEnter` and `onMouseLeave` in `resolve-interaction-styles-plugin.js` and found that when the hover style was not being removed -`e.target` had the value of the next element that was entered, not the value that was left.

So this doesn't seem like the fault of radium, but it seems like something you may need to deal with! I would expect `:hover` to be a unique property that can only exist on a single element - it should not be able to get into a state where two different elements have the hover state at the same time. 

I guess this implies some sort of global state than ensures this?

wbecker:
Fixes #367 

I haven't written tests around this, as I'm not sure you will accept it due to the (admittedly ugly!) global state. 

However, this has solved my problem, when running it locally. 

If this is accepted in theory, I'll make it ready for pulling!

### DIFF
--- a/modules/plugins/resolve-interaction-styles-plugin.js
+++ b/modules/plugins/resolve-interaction-styles-plugin.js
@@ -10,6 +10,8 @@ var _isInteractiveStyleField = function (styleFieldName) {
     styleFieldName === ':focus';
 };
 
+var lastHover;
+
 var resolveInteractionStyles = function (config: PluginConfig): PluginResult {
   var {
     ExecutionEnvironment,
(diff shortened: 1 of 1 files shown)


ianobermiller:
The one thing this doesn't cover is nested elements, where it is certainly possible to have :hover on multiple elements at once.

wbecker:
Good point!

ianobermiller:
Closing this PR for now, feel free to reopen if you can think of a better approach.

Issue #373, opened 2015-10-01
Allow changing prefixer for Style and keyframes


joseph887:
Is it possible to somehow configure prefixing in the ```<Style>``` component for server-side rendering? Unless I missed something in the docs, It doesn't look like there is a way to use the recently-added plugins configuration for the ```<Style>``` component.

ianobermiller:
That's correct, plugins are currently for the main Radium function only. We
need to at least expose the prefixer function.

PR #379, opened 2015-10-11
Allow replacing prefixer in keyframes and <Style/>

ianobermiller:
- Keyframes gets its own feature detection and prefixing from MDN
- Prefixer can be simplified, since both keyframes and Style do the
dash case conversion themselves

### DIFF
--- a/docs/api/README.md
+++ b/docs/api/README.md
@@ -222,9 +222,9 @@ function ConfiguredRadium(component) {
 class MyComponent extends React.Component { ... }
 ```
 
-You will typically want to put plugins before the final `checkProps` so that you can still benefit from the checks it provides. If your plugin might produce other pseudo-style blocks, like `@media` consumed by `resolveMediaQueries` or `:hover` consumed by `resolveInteractionStyles`, you would want to have your plugin run before those plugins. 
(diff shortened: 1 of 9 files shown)


docs/api/README.md
```diff
@@ -249,11 +249,13 @@ Usage:
 
 ## keyframes
 
-**Radium.keyframes(keyframes, componentName)**
+**Radium.keyframes(keyframes, componentName, prefix)**
```
AnSavvides: What are your thoughts on using the standard `(required, [optional])` pattern to denote whether a parameter is optional or not?

modules/keyframes.js
```diff
@@ -1,33 +1,47 @@
...
 
 var ExecutionEnvironment = require('exenv');
 
-var isAnimationSupported = ExecutionEnvironment.canUseDOM &&
-  Prefixer.getPrefixedPropertyName('animation') !== false;
+var isAnimationSupported = false;
+var keyframesPrefixed = 'keyframes';
+
+if (ExecutionEnvironment.canUseDOM) {
+  // https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Animations/Detecting_CSS_animation_support
```
AnSavvides: A short blurb might be nice to go along with the URL :+1:

ianobermiller:
Also fixes #373

PR #411, opened 2015-11-11
add active fix

### LINKED ISSUE #410: Focus is getting stuck, but only on one element
I have two buttons side by side and both use the same focus styling. For some strange reason, all buttons but the first one get "stuck" in focus. For some reason the event doesn't fire to remove it ever.

Here's a video:

https://v.usetapes.com/AIS3nzVHbg

Any ideas why this would be happening?

natew:
Ignore the flint-radium, but this fixed the bug for me.

Fixes #410

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,5 +1,5 @@
 {
-  "name": "radium",
+  "name": "flint-radium",
   "version": "0.14.3",
   "description": "A set of tools to manage inline styles on React elements",
   "main": "lib/index.js",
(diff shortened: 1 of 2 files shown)


ianobermiller:
See my comment in #410 on the correct fix.

natew:
@ianobermiller would you want me to patch that or do you have it?

ianobermiller:
Oh shoot, I should have fixed this before releasing 0.15.0 10 minutes ago. I'll fix it right now and put out 0.15.1

Issue #413, opened 2015-11-12
Wrong -ms vendor prefix


PallasKatze:
Prefixer yields `ms-transform` instead of `-ms-transform` in IE10

PR #414, opened 2015-11-12
Fix -ms vendor prefix

PallasKatze:
Closes FormidableLabs/radium#413


ianobermiller:
Thanks for contributing!

PR #418, opened 2015-11-14
Call getChildContext() instead of returning it

### LINKED ISSUE #412: getChildContext returning function
I assume this is a typo.
The getChildContext method of the RadiumEnchancer returns a function instead of an actual Object.
a simple function call fixes this.
```javascript
const superChildContext = super.getChildContext ?
        super.getChildContext :
        {};
```
to
```javascript
const superChildContext = super.getChildContext ?
        super.getChildContext() :
        {};
```

notice the function call on super.getChildContext if it exists.

richardfickling:
fixes #412

### DIFF
--- a/src/enhancer.js
+++ b/src/enhancer.js
@@ -91,7 +91,7 @@ export default function enhanceWithRadium(
 
     getChildContext() {
       const superChildContext = super.getChildContext ?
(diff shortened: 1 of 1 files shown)


ianobermiller:
Thanks for the fix!

PR #423, opened 2015-11-17
Keyframes Server Side Rendering

ianobermiller:
Allow keyframes to be rendered correctly on the server. Uses React to render a style sheet with the keyframes.

Requires you to mark your App component with `isRoot`, and causes the rendered element from your root App component to be wrapped in a div, so a style element after.

Exposes a new method to plugins, `addCSS` which causes a chunk of CSS to be added to the aforementioned style shset.

Because the style sheet component appears after the user’s rendered elements, the chunks are added correctly during a server-side render.

A StyleKeeper is created in the root component and passed down via context to the children, who can add CSS as they want.

Keyframes is now a combination of the `Radium.keyframes` helper, which creates an opaque object that must be assigned to the `animationName` property. The plugin, `Radium.Plugins.keyframes`, detects this special object, calls `addCSS`, and transforms the value into the actual animation name.

This same mechanism will also be used to add media query styles.

### DIFF
--- a/.flowconfig
+++ b/.flowconfig
@@ -20,3 +20,6 @@ src
 
 [libs]
 interfaces
+
+[options]
+suppress_comment= \\(.\\|\n\\)*\\$FlowStaticPropertyWarning
--- a/docs/api/README.md
+++ b/docs/api/README.md
@@ -4,6 +4,7 @@
 
 - [Sample Style Object](#sample-style-object)
 - [Radium](#radium)
+  - [config.isRoot](#configisroot)
   - [config.matchMedia](#configmatchmedia)
   - [config.plugins](#configplugins)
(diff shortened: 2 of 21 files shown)


coopy:
@ianobermiller what do you need in order to land this PR? Let me know if I can help!

ianobermiller:
@coopy Mostly just validation that this is an approach we want to take. it requires you to have radium with the isRoot config on your App component. I'm not sure this is the best way, and I don't want to churn too much. However, Radium really needs server rendering, so it is probably worth getting something working!

ianobermiller:
Alright I just pushed a rebased version. Now I just need to rewrite the keyframes tests and this is good to go.

ianobermiller:
:shipit:

PR #437, opened 2015-12-01
Correct and clarify media query syntax

mars:
Based on real-world usage of version 0.15.3, including server-side rendering.

* style objects need camel-cased selectors, prefixed with `@media`
* `<Style/>` components need dasherized selectors, nested in a `mediaQueries` key


mars:
Opened prematurely.

PR #440, opened 2015-12-02
Make animation name more unique

craigbilner:
To make the animation name more environmentally friendly (so that it doesn't clash with external css) use a radium specific name "-radium-animation-x-y"
To stop animation names clashing between radium versions (a dependency which depends on a different radium package and therefore has it's own radium) seed the name with the number of style sheets already appended to head "x" with the number of current animations as before "y"

The name could actually be anything including a hash of the rule or "document.head.childElementCount" for example, but this seemed to be the most elegant way with little overhead.

Had to change the tests because the sinon stubs were misbehaving and interfering with the original methods even after restoration. Less DRY but makes test much more isolated.

### DIFF
--- a/src/__tests__/keyframes-test.js
+++ b/src/__tests__/keyframes-test.js
@@ -1,9 +1,11 @@
 let styleElement;
 let exenv;
+let sandbox;
 
 describe('keyframes', () => {
 
   beforeEach(() => {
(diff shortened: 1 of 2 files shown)


ianobermiller:
LGTM, much prefer the sinon sandbox, nice work.

PR #447, opened 2015-12-09
Do not add @media into inline styles

ondrejbartas:
This will fix problem with inserting @media on server side

screenshot 2015-12-09 16 46 26


ianobermiller:
The correct place to fix this is in the media query plugin; it should strip keys starting with @media before it bails. I have a patch for this already, I'll send out a PR soon.

ianobermiller:
Thanks for bringing this up, and for the PR!

ondrejbartas:
@ianobermiller thx

PR #454, opened 2015-12-14
Move dev dependencies to `devDependencies`

herrstucki:
I ran into the problem that I couldn't use Radium in a Babel 6 app because `babel@5.8.23` was in Radium's dependencies. I also moved some other dev dependencies which weren't used anywhere in `src/` to `devDependencies`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,17 +33,16 @@
   },
   "license": "MIT",
   "dependencies": {
-    "array-find": "^1.0.0",
-    "babel": "^5.8.23",
-    "babel-plugin-flow-comments": "^1.0.9",
     "exenv": "^1.2.0",
     "inline-style-prefixer": "^0.5.3",
-    "is-plain-object": "^2.0.1",
-    "rimraf": "^2.4.0"
+    "is-plain-object": "^2.0.1"
   },
   "devDependencies": {
+    "array-find": "^1.0.0",
+    "babel": "^5.8.23",
     "babel-eslint": "^4.1.1",
     "babel-loader": "^5.3.2",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -33,17 +33,16 @@
...
-    "babel-plugin-flow-comments": "^1.0.9",
     "exenv": "^1.2.0",
     "inline-style-prefixer": "^0.5.3",
-    "is-plain-object": "^2.0.1",
-    "rimraf": "^2.4.0"
+    "is-plain-object": "^2.0.1"
   },
   "devDependencies": {
+    "array-find": "^1.0.0",
+    "babel": "^5.8.23",
```
ryan-roemer: Used by `postinstall`

package.json
```diff
@@ -73,6 +72,7 @@
     "react": "^0.14.0",
     "react-addons-test-utils": "^0.14.0",
     "react-dom": "^0.14.0",
+    "rimraf": "^2.4.0",
```
ryan-roemer: Used by `postinstall`

herrstucki:
Ah yes, I see that they're used in `postinstall`. Thanks for pointing out @ryan-roemer!

But I wonder why `postinstall` is even needed, `lib/` is built during `prepublish`? In any case the `babel@5` dependency breaks apps which are built with Babel 6.

ryan-roemer:
@herrstucki -- `lib/` is not in source, so if you do a dependency like:

```js
"radium": "FormidableLabs/radium"
```

or off of a fork / working git branch / etc. you'll need to _create_ `lib/`.

Issue #323, opened 2015-08-19
Standard bootstrap CSS file?


bbirand:
I'm trying to wrap my head around Radium, and whether it would make sense for me to transition to it. I currently am maintaining a site that uses the Bootstrap template. I really do not want to write the entire bootstrap codebase using the object notation. I found the `css-to-radium` project, but converting the `bootstrap.css` file using that project results in an error.

This made me wonder; am I even going about this in the right way? Given the popularity of Bootstrap, I would have imagined there would be a "Radium-ready Bootstrap" that I can directly `import`. Is there a better way of doing this?

PS: I found the radium-bootstrap project, but that seems to have a different aim (emulating Bootstrap functionality using React components...).

AnSavvides:
@bbirand I think it's worth making this distinction, even though it might seem obvious:
- Radium enables you to write inline styles, so it's just a tool to enable you to write your styles in a particular way
- Bootstrap is a set of already existing styles that you can access and make use of by specifying certain classes

As far as I can tell, I think if you need the power of Bootstrap then you should just use `bootstrap.css` and specify the classes you need within your React components (as you are doing now, I assume). Then the rest of your styles, outside of the realm of Bootstrap, if you wish for them to be inline-only just use Radium.

Here is a simple example where you create a submit button that utilizes Bootstrap (via `className="btn btn-primary"`) but also makes use of `Radium` for extra styling and hover effects:

```js
"use strict";

import React from "react";
import Radium from "radium";

@Radium
export default class RadiumWithBootstrap extends React.Component {
    render() {
        const style = {
            color: "pink",
            ":hover": {
                color: "yellow"
            }
        };

        return (
            <div>
                <input
                    type="submit"
                    value="Lookup"
                    className="btn btn-primary"
                    style={style}
                />
            </div>
        );
    }
}
```

alexlande:
I think @AnSavvides is right on the money here. Radium works very nicely in conjunction with an existing CSS framework. You can write your application-specific styles with Radium without any risk of naming collisions or specificity issues with the framework styles.

bbirand:
I see, so the usage pattern is to include bootstrap in the root HTML, and to style the components as necessary via Radium. This seems like a fair compromise. The components are not fully self-sufficient, but I guess depending on bootstrap is not that bad. Thank you both for your replies!

PR #469, opened 2015-12-23
fixed docs highlighting and refctored a bit

knowbody:


knowbody:
please wait before merging this

ianobermiller:
I was using `as` instead of `js` because it has better highlighting for JSX. It probably makes sense to only use it when we are using JSX, though, which isn't that often.

docs/api/README.md
```diff
@@ -17,7 +17,7 @@
 
 ## Sample Style Object
 
-```as
```
alexlande: The `as` thing is a hack for improved JSX support (otherwise syntax highlighting gets broken whenever it encounters any JSX in `js` mode) (unless they fixed that, in which case never mind!)

knowbody:
jsx is better in those places where we use jsx it highlights okay now, fixing the PR

docs/api/README.md
```diff
@@ -489,7 +489,7 @@ In order to fully support print styling it is necessary to use CSS because of br
 
 With ES7 decorators and static class properties:
 
-```as
+```js
```
ianobermiller: highlighting looks really bad in this block without JSX support

knowbody: give me 2secs, sending the fix now

knowbody: hope I won't fail :stuck_out_tongue_winking_eye:

docs/faq/README.md
```diff
@@ -144,20 +144,22 @@ You can find a more advanced example in the [react-styling readme](https://githu
 
 ## Can I use Radium with Bootstrap?
 
-See https://github.com/FormidableLabs/radium/issues/323 for discussion.
+See [issue#323](https://github.com/FormidableLabs/radium/issues/323) for discussion.
```
ianobermiller: nit: `issue [#323](...)`

docs/faq/README.md
```diff
@@ -144,20 +144,22 @@ You can find a more advanced example in the [react-styling readme](https://githu
...
-```as
+
+```js
 var Link = require('react-router').Link;
 Link = Radium(Link);
 ```
-Huge thanks to @mairh for coming up with this idea in https://github.com/FormidableLabs/radium/issues/324.
 
-We are also exploring adding a mechanism to bypass Radium's check, see https://github.com/FormidableLabs/radium/issues/258.
+Huge thanks to [@mairh](https://github.com/mairh) for coming up with this idea in [issue#324](https://github.com/FormidableLabs/radium/issues/324).
```
ianobermiller: nit: `issue` outside of link

docs/faq/README.md
```diff
@@ -144,20 +144,22 @@ You can find a more advanced example in the [react-styling readme](https://githu
...
+
+We are also exploring adding a mechanism to bypass Radium's check, see [issue#258](https://github.com/FormidableLabs/radium/issues/258).
```
ianobermiller: nit: `issue` outside of link

knowbody:
@ianobermiller check the highlighting now

alexlande:
I had no idea GitHub added `jsx` highlighting support. That's awesome!

docs/faq/README.md
knowbody: like: issue[#123](https://github.com/123) ?

ianobermiller:
Looks much better, that is awesome. Since I assume jsx is a superset of JS, can we just always use jsx? That makes it easier to remember.

knowbody:
yeah absolutely, I will fix it and squash the commits

knowbody:
@ianobermiller what about those issue hyperlinks?

ianobermiller:
Let's either only link the `#123` part, or put a space between the `#` and `issue`.

knowbody:
:+1: cool, thanks

knowbody:
@ianobermiller all done now

PR #473, opened 2015-12-24
Media Query SSR

### LINKED ISSUE #466: Media Query Server Side Rendering (SSR)
Now that the keyframes PR #423 has landed, we have the capability to implement media queries on the server side. This is basically a brain dump of my implementation plan.

SSR of media queries using Radium will be a hybrid approach. The top level css properties in the media query style (i.e. no modifiers) will be compiled to CSS and served as a style sheet in the root component. The styles nested inside a media query will be processed in JavaScript as they currently are (basically just added to the style object if the media query matches, to be processed by the next plugin).

As an example of the split, if you have this inline style object:

```js
{
  '@media (...)': {
    width: '90%',
    ':hover': {
      color: 'black'
    }
  }
}
```

the 'width` property would be compiled to CSS, and the `color` property in the `:hover` block would be applied using JS as it is now.

Here are the high level steps:

1) When [merging styles in the `resolveMediaQueries` plugin](https://github.com/FormidableLabs/radium/blob/28503741f5ab09f40d3bd821bf8147a2d16c6360/src/plugins/resolve-media-queries-plugin.js#L77), don't merge in top level styles. Before merging in any nested styles, add `!important` to each rule (to override the one in step 2)
2) Instead, compile the top level styles to CSS via `cssRuleSetToString`. You'll have to add `!important` to each rule to override other possible inline styles.
3) Generate a class name for the styles (a hash of the rule set is probably fine, you can use the hash function from keyframes) and build the full `@media ...` css
4) In the plugin, call `addCSS` to inject the CSS into the document
5) Also in the plugin, add the generated class name to the elements props

ianobermiller:
Fixes #466, more details in the task. Had a few spare minutes, so I whipped this up for you. Merry Christmas :)


knowbody:
@ianobermiller you rock!

isTravis:
Awesome! Along with #447 - I'm pumped for the next release!

ads1018:
Great to see Radium will be supporting SSR media queries. Hopefully this quells the last vestiges of skepticisms some developers have towards inline styles. I'm looking forward to remedying the checksum error I'm being thrown in my universal React app. Any idea when this feature will be made available?

ads1018:
Tested this branch on my machine. This command fails on my system: `rimraf lib && babel src/ -d lib/`
[npm-debug.log.zip](https://github.com/FormidableLabs/radium/files/76348/npm-debug.log.zip)

coopy:
Yay! `\o/`
Thank you, @ianobermiller! Happy new year :]

PR #478, opened 2016-01-03
Add StyleRoot component to replace isRoot config

ianobermiller:
`isRoot` had the odd behavior that it wrapped its contents in a div. Instead, make the “style root” a separate, explicit component.

Definitely open to feedback about the name!

### DIFF
--- a/src/__tests__/media-query-test.js
+++ b/src/__tests__/media-query-test.js
@@ -1,6 +1,6 @@
 /* eslint-disable react/prop-types */
 
-import Radium from 'index.js';
+import Radium, {StyleRoot} from 'index';
 import React, {Component} from 'react';
 import ReactDOM from 'react-dom';
 import TestUtils from 'react-addons-test-utils';
@@ -199,8 +199,9 @@ describe('Media query tests', () => {
       matches: true
     }));
 
-    @Radium({isRoot: true, matchMedia})
-    class TestComponent extends Component {
+
+    @Radium
+    class ChildComponent extends Component {
       render() {
         return (
           <span style={{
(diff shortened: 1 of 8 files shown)


ianobermiller:
This also has another difference from `isRoot`: you can't have keyframes or media queries in the same component that renders `StyleRoot`, e.g.:

```jsx
render() {
  return (
    <StyleRoot>
        <div style={{'@media (min-width: 100px)': {color: 'red'}}} />
    </StyleRoot>
  );
}
```

This is unlikely to be a problem in real apps, but makes the tests slightly annoying.

ianobermiller:
Another alternative is that any Radium component will create the StyleKeeper if it doesn't receive one in context, and you must render the `<StyleSheet>` component manually:

```jsx
import Radium, {StyleSheet as RadiumStyleSheet} from 'radium';

@Radium
class App extends Component {
  render (
    <div>
      ...rest of your elements...
      <RadiumStyleSheet />
    </div>
  );
}
```

The downside to this approach is that the failure case is not detectable automatically; we can't warn you if you forget to render StyleSheet.

ianobermiller:
cc @knowbody @alexlande for feedback

bobbyrenwick:
I like the idea of warning when you're not using Radium correctly, explicit failures are going to ease new Radium users into using the library.

I also think that the `StyleRoot` component looks good, and the docs explain that it renders a `div` and accepts all the props that a `div` would anyway.

Just a question about client side rendering with this new approach. There's no performance hit in terms of recalculating styles from calling `addCSS` whenever a media query / keyframe is come across during the render of an enhanced component right?

alexlande:
I like the concept of an explicit `<StyleRoot>` component over `<RadiumStyleSheet />` with an implicit style root. A container component feels nice and understandable, with less room for human error.

src/__tests__/media-query-test.js
```diff
@@ -210,15 +211,27 @@ describe('Media query tests', () => {
       }
     }
 
+    @Radium({matchMedia})
+    class TestComponent extends Component {
```
ianobermiller: @bobbyrenwick @alexlande See how I had to add an extra wrapper component here? Is this an acceptable tradeoff for `StyleRoot`? Just want to mention it explicitly, since you are both in favor of this approach.

alexlande: I think so, yeah :+1:

bobbyrenwick: I think this is fine.

Couldn't this be abstracted into a function so that all future tests requiring a StyleRoot would get wrapped automatically? Or is that not possible with TestUtils?

ianobermiller: Sure, this could certainly be abstracted for tests. My preference, though, is for test to be as self-contained as possible.

PR #479, opened 2016-01-03
Remove printStyles feature

ianobermiller:
No longer needed, since media query styles are now rendered as CSS
anyway.

cc @bobbyrenwick


bobbyrenwick:
I'm a little concerned that after this there's no tests ensuring that print media query styles get extracted into the top level `isRoot` or soon to be (`<StyleRoot>`) component. How about adding a print media query test like the one added in https://github.com/FormidableLabs/radium/commit/cb58b5dfcb2f9ba188906e55e6f03a8fb727a9c7?

bobbyrenwick:
I should also have said that the fact that SSR and print styles now have the same implementation is great. I'm glad to see this go and the API surface area become smaller!

ianobermiller:
Yes I meant to add a test, thanks for the reminder.

PR #499, opened 2016-01-07
Fix two minor bugs

ads1018:
- Bind the private method `_onChange()` to the class instance in `stylesheet.js`. This does, however, introduce a new bug if you're using react-router but I couldn't get to the bottom of it. The error reads: `Cannot update during an existing state transition`.
- Add missing semicolon to static member (caused an error) `inside radium-test.js`

### DIFF
--- a/src/__tests__/radium-test.js
+++ b/src/__tests__/radium-test.js
@@ -287,7 +287,7 @@ describe('Radium blackbox tests', () => {
   it('resolves styles if an element has element children and spreads props', () => {
     @Radium
     class Inner extends Component {
-      static propTypes = { children: PropTypes.node }
+      static propTypes = { children: PropTypes.node };
       render() {
         return (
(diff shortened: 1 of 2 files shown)


ianobermiller:
Cool, thank you! I should have caught that with style-sheet. Could you possibly add a regression test as well?

ianobermiller:
Going to merge this now, since this does appear to be broken :S

The semicolon is actually not required by the spec, not sure why babel is barfing on it.

ads1018:
Fantastic. I created another pull request with your suggestion to wrap setState inside a timeout and keep to track of isMounted. I'd be happy to write the regression test though to be honest I haven't written one before :S

ianobermiller:
The tests are pretty simple, check out radium-test.js, or the keyframes or media query tests, which are much smaller.

ads1018:
Great, will do. Thanks!

PR #501, opened 2016-01-07
Resolve warnings caused by StyleSheet

### LINKED ISSUE #500: Error while using StyleRoot in v0.16.1 with React Router
I've wrapped my top level component in StyleRoot and it works great! However, when transitioning to a new route (I'm using React Router) Radium throws this error:
`Cannot read property 'setState' of undefined`. 

It seems the function `_onChange()` on line 36 in `style-sheet.js` is scoped to the wrong `this`. Adding the following to the constructor fixes this bug:
`this._onChange = this._onChange.bind(this);` I submitted a pull request for this here #499. 

Fixing this does however introduce a warning when transitioning to a new route: `Warning: setState(...): Cannot update during an existing state transition (such as within `render`). Render methods should be a pure function of props and state.` Any ideas what's causing that?

ads1018:
This wraps setState inside a setTimeout and keeps track of isMounted to avoid warnings referenced here -> #500 

Fixes #500

### DIFF
--- a/src/components/style-sheet.js
+++ b/src/components/style-sheet.js
@@ -25,9 +25,14 @@ export default class StyleSheet extends Component {
   }
 
   componentWillUnmount() {
+    this._isMounted = true;
     if (this._subscription) {
       this._subscription.remove();
     }
+  }
+
+  componentWillUnmount() {
+    this._isMounted = false;
   }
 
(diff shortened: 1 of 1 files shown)


ianobermiller:
image

ads1018:
high-five-cat_o_646881

PR #531, opened 2016-01-16
Do not reprocess children's style if they are Radium-enhanced components themselves

spacenick:
Hey guys, at try.com we're using Radium on many low-level components (namely, a View wrapper which is a kind of smart component around div, adding flex capabilities and many other things), and we've noticed poor performances when there's successive re-renders (even though all of our components are Pure and we use immutable flux stores)

Our tree consists of a lot of View inside other View (makes sense right, since this is the equivalent of div for us), and so, I discovered that the way the mapping of the children is done atm is making these components re-process their styles, even though they were going to process it anyway because they're all Radium-enhanced components.

This commit makes dramatic improvements in our frame rate.

Does that make sense? Is there something I'm not seeing/not thinking of?

Cheers! We love Radium :)

### DIFF
--- a/src/resolve-styles.js
+++ b/src/resolve-styles.js
@@ -31,8 +31,13 @@ let resolveStyles = ((null: any): (
   component: any, // ReactComponent, flow+eslint complaining
   renderedElement: any,
   config: Config,
-  existingKeyMap?: {[key: string]: bool}
+  existingKeyMap?: {[key: string]: bool},
+  shouldResolveCheck: true
 ) => any);
+
+const _shouldResolveStyles = function(component) {
+  return (component.type && !component.type._isRadiumEnhanced);
+};
 
 const _resolveChildren = function({
   children,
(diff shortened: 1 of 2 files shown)


src/resolve-styles.js
```diff
@@ -62,17 +62,26 @@ const _resolveChildren = function({
     };
   }
 
+  const shouldResolveStyles = function(element) {
```
ianobermiller: prefix with `_` and move up one scope above `_resolveChildren`

src/resolve-styles.js
```diff
@@ -62,17 +62,26 @@ const _resolveChildren = function({
...
+    return (element.type && !element.type.__isRadiumEnhanced);
+  };
+
   if (React.Children.count(children) === 1 && children.type) {
     // If a React Element is an only child, don't wrap it in an array for
     // React.Children.map() for React.Children.only() compatibility.
     const onlyChild = React.Children.only(children);
-    return resolveStyles(component, onlyChild, config, existingKeyMap);
+
+    if (shouldResolveStyles(onlyChild)) {
```
ianobermiller: nice! We should also add this every else we recursively call resolveStyles; alternatively, add it to resolveStyles itself, add a flag to skip the check, and only skip the check in the exported function.

ianobermiller:
Looks great, and I'll bet this will help with the other double-processing we saw, which is why we added `_radiumDidResolveStyles` in the first place. Please add the check in a few more places.

If you can figure out how to add a unit test that would be perfect. But, since the [`_runPlugins` already ignores custom components](https://github.com/spacenick/radium/blob/b75b445f08ea0de6c46ae19de61d21707cb4cf3f/src/resolve-styles.js#L199), there probably won't be a testable side effect, so if you can't, that is fine too.

spacenick:
Cool, I'm gonna work on the edits you mentioned then!

ianobermiller:
@spacenick Yeah absolutely, `_radiumDidResolveStyles` is at a different level, I like this approach much more. It remains to be seen whether or not we can remove `_radiumDidResolveStyles`.

ianobermiller:
Thanks for updating, looks good to merge!

spacenick:
Thanks to you! Are you gonna bump npm too? Thanks!

arush:
would love to see this on npm too ;)

chriswalkr:
Any idea when this will make it to NPM?

Issue #546, opened 2016-01-21
Autoprefix for all user agent


oliviertassinari:
I'm working on the server-side rendering, my `renderToString()` is taking too much time (100 ms +).
So I have added some memoization. This memoization is taking the `user-agent` as a parameter.

Is it possible making radium apply the prefixes for a large number of browsers?
I would be able to remove the `user-agent` as a parameter of the memoization. This would increase my `index.html` size, but reduce the memory usage of my server. That's a trade-off I'm willing to take.

ianobermiller:
Yes, this is possible, as `inline-style-prefixer` does support it. `inline-style-prefixer` added a special API for this, `prefixAll`, and Radium could call that API if, say, you pass `all` as the `userAgent`.

oliviertassinari:
Alright, so I'm not mistaking, that's something that's currently not supported by radium but you would accept a pull request to address it. I also have to address it at the `material-ui` level.
I will probably submit a PR.

ianobermiller:
Yes, I'd take a PR for this.

doctyper:
@oliviertassinari have you seen [radium-plugin-prefix-all](https://github.com/rofrischmann/radium-plugin-prefix-all)? This is what I use to solve this use-case.

oliviertassinari:
> The plugin, while nice for basic cases, is not quite ideal, because the prefixer isn't fully pluggable. It is used directly by keyframes, style, and media queries, so none of those would be affected by swapping the plugin.

PR #552, opened 2016-01-24
[Prefix] Add an option to prefix for all user agent

oliviertassinari:
My use case for this, is to prefix for all user agent on the server side
and only for the browser agent on the client side. As pointed out here https://github.com/facebook/react/issues/1979
I think that it's a very good trade-off.
This allow me to remove the user agent from the memoization I use on the `renderToString` method.

The test fails because of https://github.com/rofrischmann/inline-style-prefixer/issues/58.
I'm gonna address it.

Fix https://github.com/FormidableLabs/radium/issues/546.

### DIFF
--- a/src/__tests__/style-component-test.js
+++ b/src/__tests__/style-component-test.js
@@ -50,6 +50,25 @@ describe('<Style> component', () => {
     `);
   });
 
+  it('add all the prefixes when we ask for', () => {
+    const output = TestUtils.renderIntoDocument(
+      <Style
+        radiumConfig={{userAgent: 'all'}}
+        rules={{div: {transform: 'rotate(90)'}}}
+      />
+    );
+
+    const style = getElement(output, 'style');
+    expectCSS(style, `
+      div{
+        transform:rotate(90);
+        -webkit-transform:rotate(90);
+        -moz-transform:rotate(90);
+        -ms-transform:rotate(90);
+      }
+    `);
+  });
+
   it('adds scopeSelector to each selector', () => {
     const output = TestUtils.renderIntoDocument(
       <Style
--- a/src/prefixer.js
+++ b/src/prefixer.js
@@ -47,7 +47,14 @@ function getPrefixer(userAgent: ?string): InlineStylePrefixer {
   }
 
   if (!_cachedPrefixer || actualUserAgent !== _lastUserAgent) {
-    _cachedPrefixer = new InlineStylePrefixer({userAgent: actualUserAgent});
+    if (actualUserAgent === 'all') {
+      _cachedPrefixer = {
+        prefix: InlineStylePrefixer.prefixAll,
+        prefixedKeyframes: 'keyframes'
+      };
+    } else {
+      _cachedPrefixer = new InlineStylePrefixer({ userAgent: actualUserAgent });
+    }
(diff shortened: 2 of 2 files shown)


oliviertassinari:
I have just realized that we can use https://github.com/rofrischmann/radium-plugin-prefix-all for the same use case. Do you still want this PR?

src/__tests__/style-component-test.js
```diff
@@ -50,6 +50,22 @@ describe('<Style> component', () => {
...
+      <Style
+        radiumConfig={{userAgent: 'all'}}
+        rules={{div: {transform: 'rotate(90)'}}}
+      />
+    );
+
+    const style = getElement(output, 'style');
+    expectCSS(style, `
+      div {
+        -ms-transform: rotate(90);
```
ianobermiller: This doesn't seem right, I would have expected `transform`, `-webkit-transform` and others to be here as well.

src/prefixer.js
```diff
@@ -47,7 +47,13 @@ function getPrefixer(userAgent: ?string): InlineStylePrefixer {
   }
 
   if (!_cachedPrefixer || actualUserAgent !== _lastUserAgent) {
-    _cachedPrefixer = new InlineStylePrefixer({userAgent: actualUserAgent});
+    if (actualUserAgent === 'all') {
+      _cachedPrefixer = {
+        prefix: InlineStylePrefixer.prefixAll,
```
ianobermiller: This will break flow (static typing), since it doesn't have all the properties of the actual [`InlineStylePrefixer`](https://github.com/FormidableLabs/radium/blob/master/interfaces/inline-style-prefixer.js).

We can definitely have a separate interface for `_cachedPrefixer`, and keep in mind that for this to be complete we also need to add all the prefixes for `prefixedKeyframes` below.

ianobermiller:
Thank you for the PR!

The plugin, while nice for basic cases, is not quite ideal, because the prefixer isn't fully pluggable. It is used directly by keyframes, style, and media queries, so none of those would be affected by swapping the plugin.

The only thing your PR is missing is support for keyframes prefix, more details in the comment.

src/__tests__/style-component-test.js
oliviertassinari: Indeed, I had no way to know the output order of the prefixes with the Symbol fail. Fixed!

src/prefixer.js
oliviertassinari: I haven't thought about the `prefixedKeyframes`. That would definitely require more work.

> That would mean changing getPrefixedKeyframes to return an array, and changing the keyframes plugin to add a chunk of style for each prefix in the array (or one big chunk with all of them, more likely.

That makes sense. I have added a `prefixedKeyframes` below to fix the `flow` check.
Can we merge the PR without the `prefixedKeyframes` support? Otherwise, I will have to submit a PR to https://github.com/rofrischmann/inline-style-prefixer first and update this PR and add some test.

ianobermiller:
We'll have to wait until rofrischmann/inline-style-prefixer#58 is fixed to merge this.

ianobermiller:
`inline-style-prefixer` has released an update, so I merged this: d0b32893c6e3e906270a51b0af00a3c4da5c8516

PR #560, opened 2016-01-26
Add TestMode

### LINKED ISSUE #529: Testing broken while using StyleRoot in v0.16
Try to update to version 0.16

I used `StyleRoot` to wrap the entire app in the top component and it works fine, but the testing was broken because the testing used shallow render in `TestUtils` module to render each single component one by one, in that way the child component did not wrap with `StyleRoot`, so the test was broken. Any way to solve this problem? Thanks!

Attached the error message;

screen shot 2016-01-13 at 12 54 08 pm

ianobermiller:
Add a `TestMode` section off of `Radium`, only accessible in non-production
builds. It has three methods:
- `clearState` - clears the global radium state, only the cache of
media query listeners at this point
- `enable` - enables TestMode, which doesn’t throw or warn as much
(right now, it just doesn’t throw when using addCSS without StyleRoot)
- `disable` - disabled TestMode

Fixes #529

### DIFF
--- a/docs/api/README.md
+++ b/docs/api/README.md
@@ -12,6 +12,7 @@
 - [Plugins](#plugins)
 - [Style Component](#style-component)
 - [StyleRoot Component](#styleroot-component)
+- [TestMode](#testmode)
 
 
 ## Sample Style Object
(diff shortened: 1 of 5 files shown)


doctyper:
@ianobermiller Looking forward to this! Any chance this PR lands soon? Looks like the only Travis failure is a lint error.

isTravis:
Also really excited for a test mode for Radium.

doctyper:
@ianobermiller Is there a reason for the delay? We are considering forking Radium, manually merging this patch in and running with that for a while, unless you are considering an alternate approach?

ianobermiller:
Nothing blocking, just haven't had any time to work on Radium lately.

PR #568, opened 2016-01-29
add .editorconfig

williamboman:
http://editorconfig.org/

I had some issues conforming to existing coding styles :smile:

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,17 @@
+# EditorConfig helps developers define and maintain consistent
+# coding styles between different editors and IDEs
+# editorconfig.org
+
+root = true
+
+[*]
+indent_style = space
+indent_size = 2
+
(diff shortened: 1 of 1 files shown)


.editorconfig
```diff
@@ -0,0 +1,17 @@
...
+[*]
+indent_style = space
+indent_size = 2
+
+end_of_line = lf
+charset = utf-8
+trim_trailing_whitespace = true
+insert_final_newline = true
+
+[*.md]
```
ianobermiller: Why would we want to keep trailing whitespace on markdown files?

ianobermiller:
Cool, thanks!

.editorconfig
williamboman: Two spaces at the end of a line  
forces a line break.

PR #569, opened 2016-01-29
examples: add spread operator tip

williamboman:
Not sure how interesting this is. I find it very useful. It's pretty heavily influenced by Sass mixins, which may be both good and bad I guess.


ianobermiller:
Looks good, thanks.

Issue #508, opened 2016-01-10
invariant violation when having custom children


almasakchabayev:
I have higher order component which accepts observable, i.e.
```
class App extends Component {
  render() {
    return <Combinator>
      {this.props.data$.map(data => <div>{data}</div>}
    </Combinator>
  }
}
```

When I wrap App with radium, i.e. Radium(App). I receive the following error:
```
Uncaught Invariant Violation: Objects are not valid as a React child (found: object with keys {source, selector}). If you meant to render a collection of children, use an array instead or wrap the object using createFragment(object) from the React add-ons. Check the render method of `App`.
```
 I know it is related to resolve-styles.js, same as with functions https://github.com/FormidableLabs/radium/issues/281

Is there a way to make Combinator work with Radium?

ianobermiller:
Right now Radium checks for a few types of children, and then passes anything that doesn't match to `ReactChildren.map`. I believe objects are still valid as children (say as the result of a previous call to `map`), but it would be worth checking. Our only other option would be to explicitly ignore observables, which looks non-trivial, unless there is a simple duck typing.

almasakchabayev:
Well, there is a way to check if children is observable by simply checking
```
if (this.props.children.subscribe) {
...
}
```

ianobermiller:
Right, cool. I'll take a look at it:

`typeof this.props.children.subscribe === function` should do the trick

almasakchabayev:
Great! Thanks

Dindaleon:
Hello,

I am having the same issue. Is there a way to fix this?

almasakchabayev:
@Dindaleon You will probably have to customize resolve_styles and send PR.

ianobermiller:
For a while now, the only thing keeping us from fixing this was that ReactChildren.map returned an opaque object. It appears to [return an array now](https://github.com/facebook/react/blob/db175052c00a65e6a852011f889c12fea50bb34b/src/isomorphic/children/ReactChildren.js#L165-L167) (the JSDocs are wrong).

So we can update resolveStyles to check if children is an array, then process it recursively, and if it is an element, process it directly, and anything else, ignore.

Dindaleon:
Nice! Will this be worked out for next release or will we have to do it ourselves as **almasakchabayev** indicated?

ianobermiller:
I don't work on Radium much at this time, so PRs are definitely welcome!

acondiff:
I am having this issue as well. Any fix for it?

[...]

PR #579, opened 2016-02-06
now works with Components that accept Observable as children

almasakchabayev:
Hi just wanted to submit PR related to this issue https://github.com/FormidableLabs/radium/issues/508

### DIFF
--- a/src/resolve-styles.js
+++ b/src/resolve-styles.js
@@ -54,7 +54,7 @@ const _resolveChildren = function({
 
   const childrenType = typeof children;
 
-  if (childrenType === 'string' || childrenType === 'number') {
+  if (childrenType === 'string' || childrenType === 'number' || children.subscribe) {
     // Don't do anything with a single primitive child
     return children;
   }


ianobermiller:
This is probably a good start. We could also get fancy and wrap the subscribe to apply Radium to any elements emitted.

src/resolve-styles.js
```diff
@@ -54,7 +54,7 @@ const _resolveChildren = function({
 
   const childrenType = typeof children;
 
-  if (childrenType === 'string' || childrenType === 'number') {
+  if (childrenType === 'string' || childrenType === 'number' || children.subscribe) {
```
ianobermiller: should we at least check that `subscribe` is a function?

ianobermiller:
`subscribe` isn't _that_ universal, it seems that each FRP library uses different naming (bacon is `onValue`, for instance). But maybe this is good enough, since many people use RxJS.

alexlande:
This one seems stalled out, going to close it. We can revisit in the future.

PR #582, opened 2016-02-10
Fix media query sorting

tptee:
In [radium-grid](https://github.com/FormidableLabs/radium-grid) we're still running into media query ordering issues. This PR appears to sort them out.

@ianobermiller would you mind looking at [this test?](https://github.com/FormidableLabs/radium/blob/master/src/__tests__/media-query-test.js#L247)
As far as I can tell, the color to expect is blue instead of white, assuming that both the 10px and 20px media queries match. Is this the right behavior?

### DIFF
--- a/package.json
+++ b/package.json
@@ -48,7 +48,8 @@
     "babel-preset-stage-1": "^6.3.13",
     "exenv": "^1.2.0",
     "inline-style-prefixer": "^0.6.2",
-    "rimraf": "^2.4.0"
+    "rimraf": "^2.4.0",
+    "sort-media-queries": "^0.2.2"
   },
   "devDependencies": {
     "babel-eslint": "^4.1.6",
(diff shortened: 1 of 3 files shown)


alexlande:
Talked with @tptee-- sorting media queries is probably not the way to go because they should be evaluated in the order that they're written (as this failing test illustrates).

It seems that the issue is that we're using objects and expecting them to have a set order when we iterate over properties, rather than using arrays. Not clear to me what a good solution is here (except to start using arrays instead of plain objects for media queries).

ianobermiller:
> It seems that the issue is that we're using objects and expecting them to have a set order when we iterate over properties, rather than using arrays

Yeah, this was a conscious choice. Object iteration order was defacto standard, and has been officially [standardized in ES6](http://www.2ality.com/2015/10/property-traversal-order-es6.html), so it is safe to rely on.

As @alexlande said, we must insert media queries in the order you list them.

Issue #575, opened 2016-02-05
Blocker: Cannot use multiple StyleRoots


Aktof:
I have an NPM react module that at its entry point instantiates the StyleRoot. I wanted to use this module in another project that uses Radium, but the two conflict with one another. In the project that uses my module, if I try to create a StyleRoot it breaks the context of the modules elements. Is there a way we can add a key to the StyleRoot to have the entities work alongside or is there a different solution.

```
Failed Context Types: Invalid context `_radiumStyleKeeper` of type `StyleKeeper` supplied to `ThemeProvider`, expected instance of `StyleKeeper`.
```

The project in question is located [here](https://github.com/aktof/react-mux).

Cottin:
I got the same issue when using a small component-library I built using Radium and using that library in another project. I did not have multiple StyleRoots but instead only one StyleRoot in my other project and no StyleRoots in the component-library (since it only exported components).

The solution that seemed to work for me was to add Radium as a **peerDependency** in my component-library instead of a normal dependency.

bobbyrenwick:
@Aktof the issue is that you have 2 different versions of `Radium` in your frontend bundle. So the `React.PropTypes.instanceOf(Radium.StyleKeeper)` check doesn't work because you have `Radium1.StyleKeeper` and `Radium2.StyleKeeper` which are in different locations in memory

If you do as @Cottin says and make `Radium` a **peerDependency** of `react-mux`, then anything that depends on `react-mux` will have to explicitly list `Radium` as a dependency. Once you've done this, you'll only have 1 `Radium` in memory and the warnings will disappear.

amavisca:
I've run into a similar problem:

I have 2 repos - one is a web app and one is a component library. The component library uses Radium and exports a bundled (webpack+babel) file. The web app also uses Radium, so when I go to render one of the components from my component library from within my webapp I get the same error:

```
Failed Context Types: Invalid context `_radiumStyleKeeper` of type `StyleKeeper` supplied to `MyComponent`, expected instance of `StyleKeeper`. Check the render method of `FooBar`.
```

I imagine peerDependencies doesn't solve this problem since I'm bundling the component library beforehand and will have a different instance of Radium. Any thoughts?

ianobermiller:
@amavisca why would you ever want to ship two copies of Radium down to the end user? It would be very wasteful, so I'm not sure why we'd want to encourage that by making StyleRoot duck-typed.

amavisca:
[...]

PR #589, opened 2016-02-15
Accept other StyleKeepers

tuckerconnelly:
Yo! Thank you for making Radium :D

I'm using Radium to create a UI library. Using the enhancer on the UI components, though, causes a problem with context. The context type is looking for an instance of its *own* package's `StyleKeeper`. However the `<StyleRoot>` in my project is passing down the project's `StyleKeeper`.

I changed the enhancer to accept the interface of the `StyleKeeper`, rather than an instance of the actual `StyleKeeper`.

Thanks! ^^

### DIFF
--- a/src/enhancer.js
+++ b/src/enhancer.js
@@ -2,7 +2,6 @@
 
 import {Component, PropTypes} from 'react';
 
-import StyleKeeper from './style-keeper.js';
 import resolveStyles from './resolve-styles.js';
 
 const KEYS_TO_IGNORE_WHEN_COPYING_PROPERTIES = [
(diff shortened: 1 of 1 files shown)


ianobermiller:
Thanks for the PR. Do we actually want to allow this, though? It encourages you to use two different version of Radium, which is error prone and very wasteful.

alexlande:
Agreed, I think the approach described in https://github.com/FormidableLabs/radium/issues/575 (make `radium` a peer dependency so that you don't have multiple versions) is the way to go unless there's some other use case this would be necessary for.

alexlande:
Thanks for the contribution! Going to close this one. Please let me know if you want to follow up on it more.

Issue #63, opened 2015-03-03
Handle (hover, active) states for touch devices.


iclanzan:
I noticed that only mouse events are taken into account when computing states. Are there any plans to support touch?

alexlande:
This is a goal, absolutely. I believe @kenwheeler had some ideas about this as well. I'd like to go over our browser state system and do some refactoring soon, so hopefully we can address this use case then.

alexlande:
See https://github.com/FormidableLabs/radium/pull/86

ianobermiller:
Is this still relevant? Since we use mouseEnter and mouseDown handlers, won't the browser automatically apply these when touching? Is the goal to support faster touches, such that we don't add those handlers to avoid the delays?

alexlande:
I think there's some benefit in applying styles based on touch events immediately, rather than relying on the simulated `:hover` and etc events handled by the browser. See discussion in #86.

ianobermiller:
From #424:

> If it detects it is on mobile, Radium should use `touchstart` and `touchend` to trigger active states. This would be very easy to add as a plugin. See https://github.com/FormidableLabs/radium/blob/master/src/plugins/resolve-interaction-styles-plugin.js for example.

goldensunliu:
I am gonna take a stab at it and see if I can get a PR

ianobermiller:
Sweet, thanks.

iclanzan:
Keep in mind that many devices (such as most Chromebooks) support both touch and mouse.

djorg83:
any update on this?

trescenzi:
Any updates on this? I've been working with React + Radium in cordova lately and would love to have a way to get the states working without a stylesheet.

Thanks!

alexlande:
We're planning on using generated CSS to handle pseudo-classes instead of emulating them ourselves. That's part of a larger effort to deal with a lot of issues, but it will resolve this one as well. See #788.

macrozone:
Also ran into this. :active state is too slow on mobile. Any updates on this?

Issue #510, opened 2016-01-11
would not it be better to support pseudo classes with style keeper just like mediaqueries and keyframes are?


pristas-peter:
suggestion

ianobermiller:
It is an interesting thought, but goes against much of what Radium was designed to do, which is to be a pure inline solution. StyleKeeper is a necessary evil, and should only be used when absolutely necessary. There are already subtle bugs with precedence and overriding once you introduce actual CSS.

pristas-peter:
I meant to still support only the hover, fpcus and active, the purpose was to let the browser do its magic behind this and remove the delay when the site bootsraps from SSR.

ianobermiller:
Yes, that would make sense. I'm also unsure of the performance implications of `addCSS` and the `StyleKeeper` modifying a style sheet. The good news is that with Radium's modular architecture, it should be trivial to create a version of the `resolveInteractiveStyles` plugin that renders to CSS.

ianobermiller:
I am actually coming around to this idea, slowly, since it would fix many bugs similar to #524 and #527. I have two concerns, though:
1. Will we be able to maintain strict overriding, such that if you have :active after :hover, it will take precedence, and so forth for media queries with nested states. If you insert the styles in the sheet in order they appear, the later ones will override the earlier ones and that will be fine. But you can imagine a case where one style is not active, so it doesn't get added, but then a later style is active, and gets added after, even though it was defined before (for example if you, for some reason, want hover to override active). Or, if we still assign classnames based on hash, an earlier component could have had the same hash (same style object) and added it to the sheet before you. Ditching deduping is fine, but how do we create a unique identifier that is still predictable for server side rendering (the answer is probably a counter in the StyleKeeper)?
2. Do we need to garbage collect styles, and will it be performant? I'll wager 99% of apps have a finite number of styles, and so not garbage collecting will be fine. But what if you actually generate a wide range of hover styles? We could remove them from the stylesheet on unmount, and since we are not deduping due to 1, we'd be OK.

As you can see there are a bunch of issues at play here. If we can do this correctly, though, this would open up just about any CSS selector to be easily used from within radium (e.g. :disabled). It also significantly changes the original architecture of Radium (pure inline styles, which has already been compromised), but really does stay true to the dynamic runtime-only awesomeness.

@alexland, @coopy, @rofrischmann, @knowbody, any thoughts?

rofrischmann:
[...]

PR #630, opened 2016-03-17
Handle (active) states for touch devices

derekcaneja:
Currently :active styles do not work on mobile because they are based on mousedown and mouseup events.
If on mobile use touchstart and touchend instead.

Fixes https://github.com/FormidableLabs/radium/issues/63

### DIFF
--- a/package.json
+++ b/package.json
@@ -48,6 +48,7 @@
     "babel-preset-stage-1": "^6.3.13",
     "exenv": "^1.2.0",
     "inline-style-prefixer": "^0.6.2",
+    "mobile-detect": "^1.3.1",
     "rimraf": "^2.4.0"
   },
   "devDependencies": {
--- a/src/plugins/mouse-up-listener.js
+++ b/src/plugins/mouse-up-listener.js
@@ -1,5 +1,9 @@
 /* @flow */
 
+import MobileDetect from 'mobile-detect';
+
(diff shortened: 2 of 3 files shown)


package.json
```diff
@@ -48,6 +48,7 @@
     "babel-preset-stage-1": "^6.3.13",
     "exenv": "^1.2.0",
     "inline-style-prefixer": "^0.6.2",
+    "mobile-detect": "^1.3.1",
```
ianobermiller: inline-style-prefixer probably already does 90% of this, so we can't afford to pull in a 35k (!) module. 

A better solution might be to always listen for touch events if they are available, and `preventDefault` inside of them so that the other handlers don't fire (the simulated mouse events)

froatsnook: Preventing default sounds dangerous.  `mobile-detect` does user-agent sniffing which is not what you want anyway.  Just do:

```javascript
const isMobile = typeof window.ontouchstart !== "undefined";
```

alexlande:
Thanks for the contribution! Going to close this one out. Please let me know if you want to reopen and keep moving forward on it. I think there's a decent chance that we'll want to solve this problem by using `addCSS` for real pseudo-selectors rather than emulated ones as in https://github.com/FormidableLabs/radium/issues/510

Issue #625, opened 2016-03-12
Overlapping media queries order can cause issues due to the use of non-unique class hashes


braposo:
This is in some way related to issue https://github.com/FormidableLabs/radium/issues/593, but I think it's not exactly the same situation so I'm creating a new issue. Feel free to mark it as duplicate if you think that's the case.

Because media queries classes aren't unique to each component, and each class is declared in the order that it's called, it can lead to some issues due to the fact that we need to use `!important` to override inline styles.

I've created a super simple example to illustrate my point, both components should be red in > 600px viewports: https://jsfiddle.net/x0fnqcnu/3/

The first `<TestingRadium>` component outputs the `md` media query, and then the second component uses the same `md` media query (which is already rendered in the code) but also outputs a new `sm` media query, which is rendered after the `md`. 

In > 600px viewports both media queries are true but the `sm` media query takes precedence over the `md` because it's interpreted after it and because we're using the `!important` it just ignores all the previous matching media queries.

I'm not sure if this is a bug at all, because it makes sense that we're able to reuse the media queries classes in multiple components, but it can cause issues as we can have multiple media queries being true but the behaviour isn't like normal CSS media queries where the last matched media query takes precedence over the others.

This happened to me when I was trying to use a responsive grid and noticed that some of the media queries were taking precedence over others and causing all sorts of layout issues. In my case I could add the `(max-width ...)` to avoid the media query overlap but I don't know if it's something that the developer needs to handle, because ultimately the behaviour is different from the CSS media queries.

Is it possible to order the media queries classes correctly or having a unique classes is the way to go?

braposo:
Just wondering if anyone had a chance of looking into this.

I was able to make a workaround in the meantime by setting `(min-width ...) and (max-width ...)`, but it requires to explicitly set the column size for every breakpoint defined in the grid, which is far from ideal. 

It's already live at https://edited.com/product/ (only the pages under product are using the new custom Radium grid) if you want a production example.

alexlande:
Hey @braposo, sorry if you already saw this, but there's a PR in progress that will resolve this issue: https://github.com/FormidableLabs/radium/pull/635

braposo:
I wasn't aware of that @alexlande, thanks for letting me know. I'll have a look tomorrow and see if there's anything worth adding to the discussion :+1:

PR #635, opened 2016-03-21
[WIP] Make media query and :visited CSS rules unique

### LINKED ISSUE #593: Media query class hashes aren't guaranteed to be unique
Steps to reproduce:
- Clone [radium-grid](https://github.com/FormidableLabs/radium-grid) and `npm install`.
- Run `builder run server-hot`.
- View the demo at localhost:3000.
- View the "Custom per-cell widths" grid.

Expected results:
Screens in order from small -> medium -> large -> xlarge.
screenshot 2016-02-16 10 30 46
screenshot 2016-02-16 10 30 55
screenshot 2016-02-16 10 31 08
screenshot 2016-02-16 10 31 19

Actual results:
Screens in same order as above.
screenshot 2016-02-16 10 33 31
screenshot 2016-02-16 10 33 22
screenshot 2016-02-16 10 33 15
screenshot 2016-02-16 10 33 04

The CSS for a working grid looks like this:
screenshot 2016-02-16 10 37 11

The CSS for the broken grid has gaps where styles disappear for some breakpoints (notice the missing 641):
screenshot 2016-02-16 10 37 28

It appears that using the media query and its styles does not guarantee uniqueness, since multiple elements can have identical media query/style combos. Therefore, StyleKeeper considers some styles as duplicates, never entering them into the cssSet.

Changing the hash to a UUID delivers the expected results without breaking any tests. Would you like a PR for this, or is there a better source of uniqueness we can derive?

tptee:
Resolves #593 and #625.

@ianobermiller @alexlande can you give this a sanity check?


src/plugins/resolve-media-queries-plugin.js
```diff
@@ -73,7 +74,7 @@ function _topLevelRulesToCSS({
     );
 
     // CSS classes cannot start with a number
-    const mediaQueryClassName = 'rmq-' + hash(query + ruleCSS);
+    const mediaQueryClassName = `rmq-${styleId()}-${hash(query + ruleCSS)}`;
```
ianobermiller: `styleId()` will execute every render. Since `addCSS` no longer dedupes due to the incrementing number, you'll end up with many duplicate styles in the CSS, right?

src/plugins/index.js
```diff
@@ -66,6 +66,9 @@ export type PluginConfig = {
   // separate from `props` for ease of use.
   style: Object,
 
+  // The StyleKeeper counter for unique style IDs.
+  styleId: () => number;
```
ianobermiller: `getStyleID` would be a better name

src/plugins/resolve-media-queries-plugin.js
tptee: Sure looks that way. The unique component id route may solve that issue, and I think it'd be a similar sized diff. Thoughts?

src/plugins/index.js
tptee: Agreed! Will change

tptee:
The approach I just pushed no longer uses a counter, but a uuid set only once on component props. This gives us deduping back and prevents coupling the plugins to the `StyleKeeper`. Thoughts?

ianobermiller:
You could also store the unique id in the component state (directly on the
instance).

tptee:
That's so much better, now it doesn't run every render and we don't have to check that we already set the  styleID on the component. Thanks for the guidance!

src/enhancer.js
```diff
@@ -63,6 +64,7 @@ export default function enhanceWithRadium(
 
       this.state = this.state || {};
       this.state._radiumStyleState = {};
+      this.state._styleID = uuid.v4();
```
ianobermiller: Do we want this to be per-instance or per-component? #625 would be solved by per-component, but #593 it seems would not be.

If it is per-instance, you should move it directly to `this`, and prefix it with radium, just like `_radiumIsMounted` below.

If it is per component class, you can make it a static, like `_isRadiumEnhanced` above.

ianobermiller:
One issue with putting a styleID on each instance is that you'd end up with a ton of bloat and duplication in the stylesheet; consider rendering a grid with 100 cells, each one adding 2-3 media queries.

tptee:
True, that would be nasty. Maybe we can change the StyleKeeper's `_cssSet` to look like `{ cssClass: rule }`. `getCSS` can then detect which classes have the same rule and generate CSS to reuse the rule like so:

```css
@media only screen and (min-width: 1024) {
  .rmq-12345, .rmq-23456 {
    display: none;
  }
}
```

I'll investigate!

tptee:
Does `_styleID` as a static on the component class avoid the stylesheet bloat? I'm fuzzy on why it would/wouldn't

src/enhancer.js
tptee: Ah right, confirmed a static _styleID doesn't solve #625

ianobermiller:
It would avoid stylesheet bloat, but might not fix usage with grid for example, where you had clashes within the same component, right?

tptee:
I tried a few approaches to reducing styleID's added bloat to the stylesheet without success. Using styleID increases the lines of CSS from 315 to 896 (~3x) in the `radium-grid` demo. However, since CSS parsing is quick (to my knowledge), this might be a worthwhile tradeoff for correctness's sake until we devise a better solution.

It may be worth reconsidering media query precedence in the future, as I've noticed designers heavily rely on the cascading behavior of `min-width` media query ranges. We could more aggressively consolidate media query rules while enforcing the usual definition-order precedence for rules within the media query.

I honestly don't know which follows the principle of least surprise: the consistency of definition-order precedence everywhere, or the cascading behavior of media query ranges that designers expect.

Thoughts? @ianobermiller @alexlande

ianobermiller:
@tptee, the 3x will be fine after gzip, but having too many selectors will definitely slow things down eventually. We need a better solution, but I am fine with this until we figure one out. The increase is linear with respect to what? The number of components rendered? If you have a grid with 100 rows, do you get 100x the style sheet size?

tptee:
I'm going to kick the tires on a more robust solution, since I don't have a clear picture of the scale at which stylesheet bloat increases. I'll hopefully have something to report back soon!

tptee:
I think I have an optimization for removing stylesheet bloat that still guarantees rule uniqueness. I've got a brain dump below, and I'd love if you all could give this a feasibility check before I continue. It's the only optimization I could find that respects the [order precedence test] (https://github.com/FormidableLabs/radium/blob/master/src/__tests__/media-query-test.js#L251).

/cc @ianobermiller @alexlande @coopy 

```css
/*

Algorithm:

- Prepare an incoming media query + ruleset using addCSS().
- Look at the previous rules defined for the same media query (if present).
- Remove rules in the current set that are duplicates of the previous set.
- Write the deduplicated current set to the stylesheet.
- Apply the class names of both the previous and current set to the component.

*/

/* Existing query */
@media only screen and (max-width: 640px) {
	.rmq-ffe912bc {
		display: flex !important;
		flex-basis: calc(50% - 8px) !important;
		align-items: flex-start !important;
		justify-content: flex-start !important;
		order: initial !important;
	}
}

/* Incoming query */
@media only screen and (max-width: 640px) {
	.rmq-69620a79 {
		display: flex !important;
		flex-basis: 100% !important;
		align-items: flex-start !important;
		justify-content: flex-start !important;
		order: initial !important;
	}
}

/* duplicates: display, align-items, justify-content, order */

/* component <---- .rmq-ffe912bc .rmq-69620a79 */
/* deduped ruleset ------> stylesheet */
@media only screen and (max-width: 640px) {
	.rmq-69620a79 {
		flex-basis: 100% !important;
	}
}

/* --------------------------------------------------------- */

/* Another incoming query */
@media only screen and (max-width: 640px) {
	.rmq-b0a85267 {
		display: flex !important;
		flex-basis: 100% !important;
		align-items: center !important;
		justify-content: flex-end !important;
		order: initial !important;
	}
}

/* duplicates from first query: display */
/* duplicates from second query: flex-basis */

/* component <---- .rmq-ffe912bc .rmq-69620a79 .rmq-b0a85267 */
/* deduped ruleset ------> stylesheet */
@media only screen and (max-width: 640px) {
	.rmq-b0a85267 {
		align-items: center !important;
		justify-content: flex-end !important;
	}
}
```

ianobermiller:
Sounds complicated, but I think it would work.

alexlande:
Yep, sounds reasonable to me.

twoam:
Any recent update regarding this PR? Cheers.

kylecesmat:
Closing this as it will not be prioritized and the branch has fallen stale

PR #645, opened 2016-03-25
Remove media query context from style component

alexlande:
In the before times, media queries were handled via a matchMedia
listening mixin at the root of apps, and queries were passed down via
context.

This bit of code stopped doing anything about a year ago.

/cc @ianobermiller @tptee


ianobermiller:
:shipit:

Issue #404, opened 2015-11-06
allow scopeSelector to list styles directly on the scoped selector


thomasjshafer:
Hey Radium,

This example doesn't seem to work:
```js
import {Style} from 'radium';
let styles = {
  display: 'block',
  mediaQueries: {
    '(min-width: 768px)':{
      display: 'none'
    }
  }
};
<span className="scope-selector">hide on desktop</span>
<Style rules={styles} scopeSelector=".scope-selector" />
```

I know the specific use-case in this example is not the best, but I wanted something easy to show. Thanks!

it produces this result: 
```css
.scope-selector display{
  0: b;
  1: l;
  2: o;
  3: c;
  4: k;
}
@media (min-width: 768px){
  .scope-selector display{
    0: n;
    1: o;
    2: n;
    3: e;
  }
}

**Comment by growlsworth** [comment]
Oh apologies it is working. I cannot scope select on the root though.

**Comment by ianobermiller** [comment]
You can just leave off the scopeSelector:

```
import {Style} from 'radium';

const styles = {
  '.scope-selector': {
    display: 'block',
  },
  mediaQueries: {
    '(min-width: 768px)': {
      '.scope-selector': {
        display: 'none'
      }
    }
  }
};

<span className="scope-selector">hide on desktop</span>
<Style rules={styles} />
```

growlsworth:
Thanks. Yeah. This turned into more of a feature suggestion than an issue. I'm enjoying Radium so far.

maximus8891:
It'd be good to support this. Looking forward to this feature.

PR #646, opened 2016-03-26
Make scopeSelector work without nested selectors

alexlande:
This updates the `Style` component to allow for applying rules directly to the `scopeSelector`, which isn't possible right now. The whole component is sort of messy and could use some love, but this is a helpful feature I think.

Resolves https://github.com/FormidableLabs/radium/issues/404

/cc @ianobermiller @tptee

### DIFF
--- a/src/__tests__/style-component-test.js
+++ b/src/__tests__/style-component-test.js
@@ -105,4 +105,32 @@ describe('<Style> component', () => {
       }
     `);
   });
+
+  it('adds scopeSelector if no selectors are present', () => {
+    const output = TestUtils.renderIntoDocument(
+      <Style
+        rules={{
+          color: 'red',
+          backgroundColor: 'white',
+          div: {
+            color: 'blue',
+            backgroundColor: 'black'
+          }
+        }}
+        scopeSelector=".scope"
+      />
+    );
+
+    const style = getElement(output, 'style');
+    expectCSS(style, `
+      .scope {
+        color: red;
+        background-color: white;
+      }
+      .scope div {
+        color: blue;
+        background-color: black;
+      }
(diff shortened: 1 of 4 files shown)


alexlande:
These spurious phantomjs install issues on appveyor are my jam :guitar: :fire:

src/components/style.js
```diff
@@ -36,6 +36,14 @@ const Style = React.createClass({
 
       if (selector === 'mediaQueries') {
         accumulator += this._buildMediaQueryString(rules);
+      } else if (typeof styles[selector] !== 'object') {
+        // If the selector is actually a rule, and not a selector,
+        // prefix it with the scopeSelector
+        accumulator += cssRuleSetToString(
```
ianobermiller: If you have several properties (say color, background, border), won't this add a separate rule with the same selector for each one?

src/__tests__/style-component-test.js
```diff
@@ -105,4 +105,28 @@ describe('<Style> component', () => {
       }
     `);
   });
+
+  it('adds scopeSelector if no selectors are present', () => {
+    const output = TestUtils.renderIntoDocument(
+      <Style
+        rules={{
+          color: 'red',
```
ianobermiller: Add a test with two props, I'd expect them to be merged under the single scope selector.

src/components/style.js
alexlande: That's embarrassing :smile:. Thanks, I'll fix this up.

alexlande:
Updated so that styles are merged into a single selector instead of using a separate one for each rule.

ianobermiller:
LGTM

PR #682, opened 2016-04-20
Fix Package Name

MicheleBertoli:
There is a typo in the comparison table (Stillr instead of Stilr).


alexlande:
Thanks @MicheleBertoli!

PR #688, opened 2016-04-26
Add & to support referencing scopeSelector

philiptzou:
Like SASS did.

### DIFF
--- a/src/components/style.js
+++ b/src/components/style.js
@@ -51,7 +51,15 @@ const Style = React.createClass({
         const completeSelector = scopeSelector
           ? selector
             .split(',')
-            .map(part => scopeSelector + ' ' + part.trim())
+            .map(part => {
+            	part = part.trim();
+            	if (part.startsWith('&')) {
+            		return scopeSelector + part.slice(1);
+            	}
+            	else {
+            		return scopeSelector + ' ' + part;
+            	}
+            })
             .join(',')
           : selector;
 


alexlande:
Could you include a use case for this? I'm not clear on what problems it would solve in Radium.

src/components/style.js
```diff
@@ -51,7 +51,15 @@ const Style = React.createClass({
         const completeSelector = scopeSelector
           ? selector
             .split(',')
-            .map(part => scopeSelector + ' ' + part.trim())
+            .map(part => {
+            	part = part.trim();
+            	if (part.startsWith('&')) {
```
alexlande: Radium supports the same browser matrix as React, so `startsWith` will need to be replaced with something more broadly supported like `part.substr(0, 1) === '&'`

alexlande:
Going to close this one until we have a concrete use case. Thanks for the contribution! Please let me know if you want to follow up on this.

philiptzou:
Sorry for the delay. Actually I bypassed this by adding an extra container.
Previously I just want to deal with tags like <div className="foo bar"
/><div className="foo nobar" />. So if Radium can support "&" operator, the
style of .foo.bar can be put into the style of .foo selector.

alexlande:
Thanks for clarifying. I'm not certain that we need this one right now, but it's definitely something we can think about in the future. I want to try to avoid making `Style` do too much.

Issue #634, opened 2016-03-18
Remove object values from inline style


imtoo:
Hi,

we are using this kind of structure for defining styles:

```js
const font = {
  fontFamily: 'Source Sans Pro, sans-serif',
  fontWeight: 'normal',
  fontSize: '14px',
  bold: {
    fontWeight: 'bold',
  },
  italic: {
    fontStyle: 'italic',
  }
}

<div style={[font, font.bold]}>bold text</div>
```

and result is working correctly but we are seeing 
![screenshot 2016-03-18 09 45 01](https://cloud.githubusercontent.com/assets/1922079/13872915/1a050a32-ecee-11e5-8012-dfa5689e5953.png)

`bold: [object Object]` would be great if radium can automatically strip those objects out.

What do you think?

braposo:
Hey @imtoo, what do you think about using something like `font.base` for the base styles instead?

```js
const font = {
  base: {
    fontFamily: 'Source Sans Pro, sans-serif',
    fontWeight: 'normal',
    fontSize: '14px',
  },
  bold: {
    fontWeight: 'bold',
  },
  italic: {
    fontStyle: 'italic',
  }
}

<div style={[font.base, font.bold]}>bold text</div>
```

This way you don't need to include the entire font object, which is causing the issue you mentioned.

Just a suggestion while I was searching for another issue :stuck_out_tongue_closed_eyes:

Rokt33r:
:+1: Not only multiple styles for one component, cascade styling also becomes better if it is possible.

```jsx
<div
  className='Workspace-info'
  style={styles.info}
>
  <div
    className='Workspace-info-userName'
    style={styles.infoUserName}
  >
  </div>
  <div
    className='Workspace-info-tags'
    style={styles.infoTags}
  >
    <TagSelect/>
  </div>
...
</div>
```
Hard to recognize how deep the cascading level of component
`this.info.user.name` vs `this.info.userName`

```jsx
<div
  className='Workspace-info'
  style={styles.info}
>
  <div
    className='Workspace-info-userName'
    style={styles['info-userName']}
  >
  </div>
  <div
    className='Workspace-info-tags'
    style={styles['info-tags']}
  >
    <TagSelect/>
  </div>
...
</div>
```
this is the less suck way

```jsx
<div
  className='Workspace-info'
  style={styles.info}
>
  <div
    className='Workspace-info-userName'
    style={styles.info.userName}
  >
  </div>
  <div
[...]

PR #703, opened 2016-05-10
Feature: remove nested style objects

jfschwarz:
This PR adds a plugin for removing nested style definitions. For some background have a look at #634 

Nested objects with a custom `toString` implementation are not removed.

### DIFF
--- a/src/plugins/remove-nested-styles-plugin.js
+++ b/src/plugins/remove-nested-styles-plugin.js
@@ -7,7 +7,7 @@ export default function removeNestedStyles({
   style
 }: PluginConfig): PluginResult { // eslint-disable-line no-shadow
   const newStyle = Object.keys(style).reduce((newStyleInProgress, key) => {
-    let value = style[key];
+    const value = style[key];
     if (!isNestedStyle(value)) {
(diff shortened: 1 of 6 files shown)


src/plugins/remove-nested-styles-plugin.js
```diff
@@ -0,0 +1,20 @@
+/** @flow */
+
+import type {PluginConfig, PluginResult} from './index';
+
+export default function removeNestedStyles({
+  isNestedStyle,
+  style
+}: PluginConfig): PluginResult { // eslint-disable-line no-shadow
+  const newStyle = Object.keys(style).reduce((newStyleInProgress, key) => {
+    let value = style[key];
```
ianobermiller: `const` instead of `let`

ianobermiller:
Looks great, thanks for your contribution!

May you please also update the docs: https://github.com/FormidableLabs/radium/tree/master/docs/api#configplugins

jfschwarz:
@ianobermiller, thanks, I did.

probedb:
Have just come across this plugin and it resolves an issue we're having. Any progress on when it'll be fixed and merged? Cheers.

PR #719, opened 2016-06-01
quote values for content property

davidjbeveridge:
When the `Style` component is used, one ought to be able to specify a rule like:
```javascript
'a:visited::before': {
  content: '√'
}
```

In master, though, this gets compiled as:

```css
a:visited::before {
  content: √;
}
```

The missing quotes prevent the `content` property from working correctly. This PR adds the missing quotes, so you'll get this:

```css
a:visited::before {
  content: "√";
}
```

### DIFF
--- a/src/css-rule-set-to-string.js
+++ b/src/css-rule-set-to-string.js
@@ -5,9 +5,14 @@ import camelCasePropsToDashCase from './camel-case-props-to-dash-case';
 import mapObject from './map-object';
 import {getPrefixedStyle} from './prefixer';
 
+function quoteValueIfNeeded(property, value) {
+  const quotedProperties = ['content'];
+  return (quotedProperties.indexOf(property) >= 0) ? `"${value}"` : value;
+}
+
 function createMarkupForStyles(style: Object): string {
   return Object.keys(style).map(property => {
-    return property + ': ' + style[property] + ';';
+    return `${property}: ${quoteValueIfNeeded(property, style[property])};`;
   }).join('\n');
 }
 
--- a/src/css-rule-set-to-string.js
+++ b/src/css-rule-set-to-string.js
@@ -6,6 +6,11 @@ import camelCasePropsToDashCase from './camel-case-props-to-dash-case';
 import mapObject from './map-object';
 import {getPrefixedStyle} from './prefixer';
 
+function quoteValueIfNeeded(property, value) {
+  const quotedProperties = ['content'];
+  return (quotedProperties.indexOf(property) >= 0) ? `"${value}"` : value;
+}
+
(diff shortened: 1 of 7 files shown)


src/css-rule-set-to-string.js
```diff
@@ -5,9 +5,14 @@ import camelCasePropsToDashCase from './camel-case-props-to-dash-case';
 import mapObject from './map-object';
 import {getPrefixedStyle} from './prefixer';
```
jmcriffey: We should probably follow convention here and make a `quote-value-if-needed` util file. Here is a similar util for reference: https://github.com/davidjbeveridge/radium/blob/2106d31209fd395ec1118478af87b4d7f06e3321/src/append-px-if-needed.js

src/css-rule-set-to-string.js
```diff
@@ -5,9 +5,14 @@ import camelCasePropsToDashCase from './camel-case-props-to-dash-case';
...
 
+function quoteValueIfNeeded(property, value) {
```
jmcriffey: Can you add in the flow types here?

src/css-rule-set-to-string.js
```diff
@@ -5,9 +5,14 @@ import camelCasePropsToDashCase from './camel-case-props-to-dash-case';
...
+  const quotedProperties = ['content'];
```
jmcriffey: Once this is pulled into its own file, this array can be defined and initialized outside of the function. That will prevent us from creating lots of array.

jmcriffey:
Looking good. I just had a few comments.

src/css-rule-set-to-string.js
davidjbeveridge: Sorry, but what do you mean by flow types?

jmcriffey: Your function definition should end up looking something like this to enable type checking with flow:

```js
function quoteValueIfNeeded(propertyName: string, value: any): string {

}
```

davidjbeveridge:
Ok, @jmcriffey I think your comments are addressed. Thanks for the feedback!

jmcriffey:
@davidjbeveridge I think you need to sync with master. It looks like your PR is diffing files already in master.

davidjbeveridge:
I've already merged in upstream/master; I think github is just not updating its diff correctly. Do you mind checking it against your local?

davidjbeveridge:
Closing, and reopening.

jmcriffey:
The diff looks good locally.

jmcriffey:
LGTM, I'll leave the final look over to @alexlande.

herrstucki:
FWIW, this breaks `content: 'none'` which before was translated to `content: none`.

Issue #727, opened 2016-06-14
Broken build


alexlande:
https://travis-ci.org/FormidableLabs/radium/jobs/137568667#L712-L715

alexlande:
Can't reproduce this locally in Node 4 or 0.10 😕

alexlande:
Resolved by #728

PR #728, opened 2016-06-14
Pin babylon 6.8.0

jmcriffey:
@alexlande @ianobermiller It looks like babel-eslint is having some trouble with babylon 6.8.1 and pinning to 6.8.0 fixes the build issue in #727 for now.

### DIFF
--- a/package.json
+++ b/package.json
@@ -46,6 +46,7 @@
     "babel-preset-es2015-loose": "^6.1.3",
     "babel-preset-react": "^6.3.13",
     "babel-preset-stage-1": "^6.3.13",
+    "babylon": "6.8.0",
     "exenv": "^1.2.0",
     "inline-style-prefixer": "^1.0.3",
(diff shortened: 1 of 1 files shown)


alexlande:
:bow: thanks! will merge this as soon as it goes green.

alexlande:
@jmcriffey: actually, one thing I'm not sure about. Should `babylon` be added to our `move-babel-to-dependencies` script? We've got a script that moves babel packages to dev dependencies on `npm publish`: https://github.com/FormidableLabs/radium/blob/master/npm-scripts/move-babel-to-dependencies.js#L6-L10

Presumably babylon should be included there too.

jmcriffey:
Good catch @alexlande. Looking at it closer, I believe I can just pin it in devDependencies since it is an issue with eslint.

alexlande:
LGTM, thanks!

Issue #581, opened 2016-02-09
BUG: `package.json:files` field breaks `git` installs.


pbrianmackey:
When I run `npm install radium --save` it errors:

> ENOENT.  This is most likely not a problem with npm itself and is related to npm not being able to 
> find a file.  Error no -4058

Looking at the attached log I saw it quit because of something found in node_modules.  So I deleted everything in node_modules and tried again.  Same error

[npm-debug.txt](https://github.com/FormidableLabs/radium/files/123667/npm-debug.txt)

pbrianmackey:
I went through the same process on a Mac and it worked.  My windows machine can't install.

tuckerconnelly:
Wasn't able to install 0.16.6 on Mac. Looks like 0.16.5 worked though :)

ianobermiller:
cc @tptee who released 0.16.6

tptee:
I can't reproduce this issue.

Tried `npm install --save radium` on the following:
- OSX 10.11, node 5.6.0, npm 3.7.0
- Windows 10, node 4 LTS, npm 2
- Windows 10, node 5.6.0, npm 3

All succeeded. Best advice I can give is to check your node and npm versions and to run a `npm cache clean`.

ryan-roemer:
Wow, that is a weird ENOENT:

```
162751 error enoent ENOENT, rename 'C:\Code\Git\ReactStylingRadium\node_modules\radium\node_modules\babel-preset-stage-1\node_modules\babel-preset-stage-2\node_modules\babel-preset-stage-3\node_modules\babel-plugin-transform-async-to-generator\node_modules\babel-helper-remap-async-to-generator\node_modules\babel-traverse\node_modules\babel-code-frame\node_modules\chalk\node_modules\strip-ansi\node_modules\ansi-regex\package.json.988ca2f102d18e457afb9bcbc99b3671'
```

Does that _look_ like a correct path for any babel-familiar folks here?

ryan-roemer:
You know, it could that windows max path thing. I heard npm 3 does a much better job with this.

@pbrianmackey -- Can you try updating to `npm install -g npm@3` and if possible update node to `4.3`+?

davidgilbertson:
I've had two strange issues.
Node 4.4
NPM 3 (so not the max path thing for me)
Windows 10
[...]

PR #730, opened 2016-06-15
Use .npmignore instead of files

jmcriffey:
@alexlande @ianobermiller This addresses #581. I decided to go the route of excluding everything except for the folders that were in the files array. If it is preferable to exclude specific files instead, let me know.


alexlande:
Nice!

I'll verify that it's working correctly this morning and merge this thing. 🔥

jmcriffey:
@alexlande I am actually still having an issue installing from git. It seems that npm still reads the .npmignore file just like it would for the files list.

jmcriffey:
So this actually solves the original issue but we still can't install from git.

ryan-roemer:
@jmcriffey -- Yeah, we may need to include `src`.

@alexlande @ianobermiller -- Is that going to be a huge deal?

alexlande:
Including `src` isn't ideal, but I don't think it would be a huge problem. It would be nice to do that for now, and then incorporate `.npmignore` into the "dirty swap on publish" scheme once that's going.

jmcriffey:
@alexlande 'll go ahead and include src in this npm ignore file for now.

ryan-roemer:
@jmcriffey @alexlande -- And we probably need `.babelrc` and maybe something else?

Side note -- we already have some of the 'dirty swap' in here: https://github.com/FormidableLabs/radium/blob/master/package.json#L29-L30 which _is_ removing `babel*` deps from package.json on publish. (See https://npmcdn.com/radium@0.17.1/package.json)

jmcriffey:
It seems like src + .babelrc allows me to install from a git repo.

alexlande:
Verified, works great :+1:

PR #732, opened 2016-06-16
Correctly quote content values in Style component

alexlande:
Github's diff getting out of sync made me nervous so I squashed this. Thanks @davidjbeveridge and @jmcriffey! Will merge once this turns green.

Resolves #719

### DIFF
--- a/src/__tests__/style-component-test.js
+++ b/src/__tests__/style-component-test.js
@@ -34,6 +34,19 @@ describe('<Style> component', () => {
     `);
   });
 
+  it('quotes properties that need it', () => {
+    const output = TestUtils.renderIntoDocument(
+      <Style rules={{'div::before': {content: '*'}}} />
+    );
+
+    const style = getElement(output, 'style');
+    expectCSS(style, `
+      div::before {
+        content: "*";
+      }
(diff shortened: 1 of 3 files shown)


alexlande:
Turns out this breaks a whole lot of other behavior around `content`, so we'll need to revert it. You'll have to explicitly quote `content` values: `content: '"Hello world"'`

davidjbeveridge:
We could introduce more logic to look for special values, eg, none, attr.

jmcriffey:
@davidjbeveridge The real issue comes with the complexities support in `content` values. For example, you could have `counter(someCounter) ". "`. That ends up combining special counter values and a string. Since it supports such custom values, you can't really get away with not quoting strings.

alexlande:
Yeah, it's unfortunate because a quoted string value is by far the most common use case for `content`, but the other factors involved make it tricky. Another plus of making you provide double-quoted values is that the same will work for any other similar properties (like font family names that need to be quoted, for example).

ianobermiller:
Ah cool. Disregard my comment on the PR. Well thought out, thanks Alex.

Issue #729, opened 2016-06-14
Update node versions in appveyor and travis


alexlande:
Right now, our builds are running in Node 0.10, 0.12, and io.js. We should probably update them to match Victory (0.12, 4.2, 5.0).

PR #733, opened 2016-06-16
Update travis and appveyor node versions

jmcriffey:
@alexlande This knocks out the small chore at #729. Do you think we need to add 5.0 for appveyor?


alexlande:
Nice! I think we're ok leaving off 5.0 for appveyor.

Issue #734, opened 2016-06-17
Switch lint warnings to errors


alexlande:
We've got a few lint rules that are warnings instead of errors. We should either have them as errors or not at all, because otherwise we end up with a bunch of lint warnings in the console that are easy to ignore.

alexlande:
Resolved by #736

PR #736, opened 2016-06-17
Change eslint warnings to errors

jmcriffey:
@alexlande I went ahead and changed all warnings to errors. We only had a small instance where warnings were given off, so it looks like all of those warning styles are usually already followed. Addresses #734.


alexlande:
Great! Thanks @jmcriffey.

PR #761, opened 2016-07-01
Fix unused property warning in StyleRoot

### LINKED ISSUE #759: Remove unknown props on <div> tag
The `style-root` component is receiving all props passed to it. We should remove unused props as ~~future versions of React~~ React v15.2.0 will throw warning messages.

```
Warning: Unknown prop `_radiumDidResolveStyles` on <div> tag. Remove this prop from the element. For details, see https://fb.me/react-unknown-prop
    in div (created by StyleRoot)
    in StyleRoot (created by App)
    in App
```

Line in question - https://github.com/FormidableLabs/radium/blob/master/src/components/style-root.js#L35

Related to this issue - https://github.com/react-bootstrap/react-bootstrap/issues/1994#issuecomment-229699699

kylecesmat:
As of React v15.2.0(https://github.com/facebook/react/pull/6800) - React issues warnings for unknown properties passed to DOM nodes. This should fix the warning when using the StyleRoot component.

I haven't tested this but it seems reasonable.

Fixes #759

### DIFF
--- a/src/components/style-root.js
+++ b/src/components/style-root.js
@@ -31,8 +31,13 @@ class StyleRoot extends Component {
   }
 
   render() {
+    /* eslint-disable no-unused-vars */
+    // Remove prop before being applied to DOM Node
+    // As of React v15.2.0 - Unknown props issue warning
+    const { _radiumDidResolveStyles, ...additionalProps } = this.props;
+    /* eslint-enable */
     return (
-      <div {...this.props}>
+      <div {...additionalProps}>
         {this.props.children}
         <StyleSheet />
       </div>


alexlande:
Thanks @kylecesmat! If appveyor ever catches up, I'll merge this thing.

conarwelsh:
+1

snakeyak:
+1

alexlande:
This doesn't do the trick, unfortunately. This prop is applied with our `_cloneElement` wrapper, so we can't handle it in the `render` method. Furthermore, this happens with *all* elements processed by Radium, not just `StyleRoot`. Taking discussion back to the open ticket to figure this out one.

Issue #762, opened 2016-07-02
Poor Stateless component detection in enhancer.js


CreepGin:
The current detection logic: 

`if (!ComposedComponent.render && !ComposedComponent.prototype.render)`

will error out when `ComposedComponent.prototype` is undefined, which will be the case when you use lambdas to define the component. 

```js
var ff = () => {}
undefined

typeof ff.prototype
"undefined"

var bb = function(){}
undefined

typeof bb.prototype
"object"
```

alexlande:
Resolved by #771. I'll plan to tag a release tomorrow morning 👍

PR #771, opened 2016-07-07
Better check for stateless component

jmcriffey:
@alexlande Addresses #762. 

As things currently stand, our stateless component detection usually works because Babel uses regular functions to compile arrows functions. This causes trouble if a user is targeting an environment that supports real arrow functions. This PR should address that case.

### DIFF
--- a/src/__tests__/radium-test.js
+++ b/src/__tests__/radium-test.js
@@ -658,6 +658,11 @@ describe('Radium blackbox tests', () => {
         {props.children}
       </div>
     );
+
+    // Babel is forced to use regular functions when defining arrow functions.
+    // Arrow functions should not technically have prototypes,
+    // so remove it here to make sure Radium doesn't fail with real arrow functions.
+    MyStatelessComponent.prototype = undefined;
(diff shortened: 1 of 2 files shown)


src/__tests__/radium-test.js
```diff
@@ -658,6 +658,11 @@ describe('Radium blackbox tests', () => {
         {props.children}
       </div>
     );
+
+    // Babel is forced to use regular functions when defining arrow functions.
+    // Arrow functions should not technically have prototypes,
+    // so remove it here to make sure Radium doesn't fail with real arrow functions.
+    MyStatelessComponent.prototype = undefined;
```
alexlande: It took me a minute to realize that this was a test and I was making all sorts of faces thinking about what kinds of damage this could do.

alexlande:
Looks great, thanks @jmcriffey!

src/__tests__/radium-test.js
jmcriffey: Haha, yeah it would be really fun chasing down bugs caused by that in the source.

PR #782, opened 2016-07-15
Replace _radiumDidResolveStyles with data-radium

### LINKED ISSUE #759: Remove unknown props on <div> tag

alexlande:
Replaces this prop with a `data-` attribute to silence unused prop deprecation warnings in React 15.2. Long term, we'll likely want to handle this in a different way that doesn't pollute the DOM like this.

Resolves #759

### DIFF
--- a/src/resolve-styles.js
+++ b/src/resolve-styles.js
@@ -286,7 +286,7 @@ const _runPlugins = function({
 const _cloneElement = function(renderedElement, newProps, newChildren) {
   // Only add flag if this is a normal DOM element
   if (typeof renderedElement.type === 'string') {
-    newProps = {...newProps, _radiumDidResolveStyles: true};
+    newProps = {...newProps, 'data-radium': true};
   }
 
   return React.cloneElement(renderedElement, newProps, newChildren);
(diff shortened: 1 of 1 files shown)


ianobermiller:
:( yeah I dislike that we will be adding to server rendered app weight;
React itself specifically eliminated data-reactid. I'll bet someone opens
an issue about these data attributes soon ;)

alexlande:
Haha, yep. Expecting that ticket any minute now.

PR #783, opened 2016-07-15
Breaking revert content quoting

### LINKED ISSUE #780: 0.17.2 takes `content: 'none'` literally unlike 0.17.1
In 0.17.1 `content: 'none'` would yielded 'none'. In 0.17.2 it seems to yield `"none"`. This is a breaking change.

Looks like it was caused by https://github.com/FormidableLabs/radium/pull/719 .

Related: https://github.com/interactivethings/catalog/issues/155 .

alexlande:
This reverts our `content` quote change because it broke behavior for keywords. Annoying, but luckily you can still use `content` with strings by explicitly double-quoting your values.

Resolves #780

### DIFF
--- a/docs/api/README.md
+++ b/docs/api/README.md
@@ -525,6 +525,11 @@ A string that any included selectors in `rules` will be appended to. Use to scop
 </div>
 ```
 
+### Notes
+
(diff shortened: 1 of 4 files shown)


ianobermiller:
Why not just whitelist keywords? It is likely rare that you you actually
want the content to be equal to a keyword, and then you can wrap it in
extra parens. Although that just results in more code in Radium, so maybe
not.

PR #835, opened 2016-09-27
Redirect homepage to /open-source

paulathevalley:
Add a redirect from http://stack.formidable.com/radium to https://formidable.com/open-source/radium. 

Resolves https://github.com/FormidableLabs/formidable-landers/issues/178


index.html
```diff
@@ -0,0 +1,56 @@
...
+    }
+    a:hover, a:focus {
+      cursor: "pointer";
+      color: #e58c7d;
+      transition: color 200ms cubic-bezier(0.22, 0.61, 0.36, 1);
+    }
+  </style>
+  <script>
+    var url = window.location.href;
+    if (url.indexOf("stack.formidable.com/radium") > -1) {
```
alexlande: Any reason not to always set `window.location`? Why is this check needed?

paulathevalley: Great question. I wrote this for victory awhile ago and can't remember. I can't imagine why it'd be needed here. Do you want to give it a go without the check?

alexlande: Sure, lets drop the check and see what happens.

alexlande:
🚢

Issue #731, opened 2016-06-15
Add `publishr` for optimal npm and git installation support.


jmcriffey:
Ideally Radium should be installable and optimized for both the npm package and the git repo. We want to be able to build and publish Radium without bogging down the main (npm) install path, but it is also advantageous to enable installing from git. Lastly, we want this process to be reliable.

We need to do a few things to accomplish those objectives:
- [x] Publish build dependencies as production dependencies in the git repo.
  - Why? You need these to `npm install` the git repo.
- [x] Publish build dependencies as devDependencies in the in the npm package.
  - Why? Babel takes a long time to install and it isn't needed for the pre-built package.
- [x] Publish `.npmignore` in the npm package.
  - Why? You save space and time if you only include the pre-built files.
- [x] Publish a different `.npmignore` in the git repo.
  - Why? The npm install command honors `.npmignore` during an install from a git repo.
- [x] Use the postversion hook to make changes.
  - Why? The prepublish hook can be unreliable. See #656.

We've been kicking around a few ideas, and I like the idea of declaring fields in `package.json` to reconcile the differences in those dependencies and ignores. We then would use a full npm version/publish lifecycle and re-write the `package.json` and `.npmignore` files depending on those shared fields.

A `package.json` file would look something like this:

``` json
{
  "dependencies": {
    "array-find": "^1.0.0",
    "babel-cli": "^6.3.15",
    "babel-core": "^6.3.15",
    "other-bulky-build-dependency": "^1.0.0"
  },
  "devDependencies": {
    "eslint": "^2.5.3"
  },
  "npmDependIgnore": [
    "^babel",
    "other-bulky-build-dependency"
  ],
  "npmFileIgnore": [
    "/*"
    "!/lib"
    "!/dist"
   ]
}
```

Our goal here to is remove `babel-cli`, `babel-core`, and `other-bulky-build-dependency` from npm's version of `package.json` and add `/*`, `!/lib`, and `!/dist` to npm's version of `.npmignore`. We could do that by using 2 new fields that dictate the differences in the git and npm repos. Our publishing lifecycle would then look like the following:
- postversion
  - Save a new `package.json` file with `npmDependIgnore` packages removed from `dependencies`.
  - Save a new `.npmignore` file with `npmFileIgnore` files added.
- postpublish
  - Revert the `package.json` file.
  - Revert the `.npmignore` file

Any thoughts or opinions relating to the above?

ryan-roemer:
Great start!

[...]

PR #841, opened 2016-10-02
let `.babelrc` not be added to package

HsuTing:


ryan-roemer:
You don't want to add `.babelrc` to `.npmignore` in the repository itself, because this will break `git`-based installations. This is because in a git-based installation, `npm pack` equivalent is run, which will remove `.npmignore`-ed files, and here the `.babelrc`, which is necessary for the `git` version to actually build `lib/` and `dist/` from scratch.

We likely want to close this PR in favor of https://github.com/FormidableLabs/radium/issues/731 which will optimally, _not_ have `.babelrc` in the `npm` registry version of Radium, but maintain it in git source.

HsuTing:
That`s great. I`m glad to hear that.

alexlande:
Closing this now that https://github.com/FormidableLabs/radium/pull/914 has landed.

ryan-roemer:
@alexlande -- We actually still do have `.babelrc` in the npm published version: https://github.com/FormidableLabs/radium/blob/master/.npmignore.publishr#L2 and https://unpkg.com/radium@0.19.3/.babelrc  Should I remove that line?

/cc @HsuTing

alexlande:
Ah, yeah, I guess so. There's no valid reason for us to have it in npm versions, right? Nothing I can think of.

ryan-roemer:
Only valid reason is if someone wanted to use https://unpkg.com/radium@0.19.3/src/ instead of https://unpkg.com/radium@0.19.3/lib/

alexlande:
Yeah, I'm good not considering that a valid use case 👍

PR #845, opened 2016-10-10
Switch StyleKeeper PropType check from instanceof to duck-typing

### LINKED ISSUE #575: Blocker: Cannot use multiple StyleRoots

Munter:
This PR Fixes #575

The use case that spawns this problem is having two different repositories,  both with instances of Radium, where one repository uses a React component from the other.

When this happens the PropType check for StyleKeeper fails on checking for identical instances, apparently because there is one instance of Radium from one repository which wants to attach components to the style root of the second repository.

This pull request adds a static proptype validator function `isStyleKeeper` to `StyleKeeper`, which compares instances using duck-typing rather than instance comparison.

The path has been verified to work when radium is linked into both repositories and components are used across them.

I have run linting, tests and flow.

### DIFF
--- a/src/components/style-sheet.js
+++ b/src/components/style-sheet.js
@@ -6,7 +6,7 @@ import StyleKeeper from '../style-keeper';
 
 export default class StyleSheet extends Component {
   static contextTypes = {
-    _radiumStyleKeeper: React.PropTypes.instanceOf(StyleKeeper)
+    _radiumStyleKeeper: StyleKeeper.isStyleKeeper
   };
 
   constructor() {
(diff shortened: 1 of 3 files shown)


Munter:
Closing this PR due to finding a better solution to #575: https://github.com/FormidableLabs/radium/issues/575#issuecomment-265699427

PR #853, opened 2016-11-08
Force Use Media Query Styles

breadbaker:
I want to be able to include a flag to promote media query styles to override top level styles.  

The purpose of this is that I am creating a wysiwyg, and I want the window of the editor to be full screen width while displaying in a smaller window.


.gitignore
```diff
@@ -1,6 +1,5 @@
 _site
 coverage
 dist
-lib
```
ianobermiller: revert, please

src/plugins/resolve-media-queries-plugin.js
```diff
@@ -74,7 +82,7 @@ function _topLevelRulesToCSS({
 
     // CSS classes cannot start with a number
     const mediaQueryClassName = 'rmq-' + hash(query + ruleCSS);
-    const css = query + '{ .' + mediaQueryClassName + ruleCSS + '}';
+    const css = _toCss(query, ruleCSS, mediaQueryClassName, props.forceMobile);
```
ianobermiller: nit: I'd rather see this function inlined, it feels noisy broken out

alexlande:
Closing this one out, please ping me if it should be re-opened.

PR #893, opened 2017-04-05
Move dependencies

christianalfoni:
I have not tested if this change works, but just wanted to bring it to attention. Services like www.webpackbin.com, and other services, analyzes package.json files and by having a lot of devDependencies in "dependencies" field it potentially breaks 3rd party stuff.

Sorry if these really are dependencies, but can not imagine that is the case? :-)

Also these dependencies takes a long time to install

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,7 +34,10 @@
     "universal": "concurrently --kill-others \"npm start\" \"npm run examples\""
   },
   "license": "MIT",
-  "dependencies": {
+  "peerDependencies": {
+    "react": "^15.3.0"
+  },
+  "devDependencies": {
     "array-find": "^1.0.0",
     "babel-cli": "^6.24.0",
     "babel-core": "^6.24.0",
@@ -46,12 +49,7 @@
     "babel-preset-stage-1": "^6.22.0",
     "exenv": "^1.2.1",
     "inline-style-prefixer": "^2.0.5",
-    "rimraf": "^2.6.1"
-  },
-  "peerDependencies": {
-    "react": "^15.3.0"
-  },
-  "devDependencies": {
+    "rimraf": "^2.6.1",
     "babel-eslint": "^7.1.1",
     "babel-loader": "^6.4.0",
     "chai": "^3.5.0",


ryan-roemer:
The babel dependencies are needed to support git-style dependencies, which needs to build `lib` not present in source, aka when you do:

```js
"raidum": "FormidableLabs/radium"
````

If you're doing a more normal `npm` install, aka:

```js
"radium": "^0.18.2"
```

The deps are slimmer as we use [`publishr`](https://github.com/FormidableLabs/publishr) to remove all the "build" dependencies and leave only the "real" dependencies.

You can see what's _actually_ published on `unpkg`: https://unpkg.com/radium@0.18.2/package.json, namely:

```js
  "dependencies": {
    "array-find": "^1.0.0",
    "exenv": "^1.2.1",
    "inline-style-prefixer": "^2.0.5",
    "rimraf": "^2.6.1"
  },
```

ryan-roemer:
Does www.webpackbin.com actually use what's in the `npm` registry? Ideally that or the services you mention like it would have an option for that to translate a project `package.json` into what's actually used on `npm`-based dependencies (as opposed to what's in git)...

If so, cool! If not, when I'm analyzing packages used in a root `package.json` via git or npm or whatever (like for https://github.com/FormidableLabs/publish-diff), I often just shell out to `npm pack` to get the real simulation of what happens at install time for the files. It's not 100%, but it's pretty close to what really happens.

christianalfoni:
@ryan-roemer Thanks for quick answer :-)

It uses the registry and I see the deps are removed there, soooo, my bad on that! Though I wonder why it takes so much time installing this package. Even when I locally run this it takes more than 1 minute to install.... you know, I would suspect it is `rimraf`, as that is a Node target.
 (using node default modules)

Any reason rimraf needs to be there related to browser?

ryan-roemer:
@christianalfoni -- Great call out for `rimraf`. I'm pretty dang sure it's not need for real dependencies. I've opened up https://github.com/FormidableLabs/radium/issues/894 for us to fix one way or another on publication.

christianalfoni:
@ryan-roemer Great stuff! It should help a lot on projects bundling stuff, as webpack/browserify has issues bundling "native modules" :-)

Thanks for following up on this! Will close this PR, pointing to: https://github.com/FormidableLabs/radium/issues/894

PR #898, opened 2017-04-28
Task use prop types

stefvhuynh:
Addresses https://github.com/FormidableLabs/radium/issues/895

PR #899, opened 2017-05-04
Import PropTypes from a separate module

iKonrad:
In React 15, importing PropTypes from the React library is obsolete and should be replaced with a separate `prop-types` module.


MichaelDeBoey:
Closed by #898

kylecesmat:
This was resolved by #898

PR #903, opened 2017-05-17
Move prop-types to `dependencies` instead of  `peerDependencies`

ryan-roemer:
- Move prop-types to `dependencies` instead of  `peerDependencies`
- Improve travis to use `yarn` for task running since it already installs using that
- Update changelog.

/cc @alexlande @mhink @MichaelDeBoey

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,7 +25,7 @@ notifications:
 cache: yarn
 
 script:
-  - npm run lint
-  - npm run test-coverage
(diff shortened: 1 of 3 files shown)


alexlande:
Looks good. Thanks for fixing the `./` paths in the scripts as well. 🚢

ryan-roemer:
Released in `radium@0.19.1`

Issue #675, opened 2016-04-17
How do I add multiple animations to keyframes?


Dindaleon:
I have this spinner:

```
const rotate = radium.keyframes({
  '100%': {
    transform: 'rotate(360deg)'
  },
}, 'rotate');
const dash = radium.keyframes({
  '0%': {
    strokeDasharray: '1,200',
    strokeDashoffset: '0'
  },
  '50%': {
    strokeDasharray: '89,200',
    strokeDashoffset: '-35'
  },
  '100%': {
    strokeDasharray: '89,200',
    strokeDashoffset: '-124'
  },
}, 'dash');
const color = radium.keyframes({
  '100%, 0%': {
    stroke: 'red'
  },
  '40%': {
    stroke: 'blue'
  },
  '60%': {
    stroke: 'green'
  },
  '80%, 90%': {
    stroke: 'yellow'
  }
}, 'color');

  loader: {
    position: 'absolute',
    width: '100px',
    height: '100px',
    left: '50%',
    top: '20%'
  },
  circular: {
    animation: 'q 2s linear infinite',
    animationName: rotate,
    height: '100px',
    position: 'relative',
    width: '100px'
  },
  path: {
    strokeDasharray: '1,200',
    strokeDashoffset: '0',
    animation: 'dash 1.5s ease-in-out infinite, color 6s ease-in-out infinite',
    animationName: dash,
    strokeLinecap: 'round'
  }
```

On the `path` class I have two animations, one for `dash` and another one for `color`
How can I handle this with radium?

techniq:
I'm also interested in how (if possible) to accomplish this.

ianobermiller:
It is currently not possible to add multiple animations.

Dindaleon:
And what would a quick fix be for this issue?

sassanh:
I'm interested too.

techniq:
@Dindaleon @sassanh - I ultimately switched to [styled-components](https://github.com/styled-components/styled-components), which supports multiple animations as well as passing in props when creating keyframes.  You can see an example I created [here](http://codepen.io/techniq/pen/JErwzX?editors=0010)

Dindaleon:
@ianobermiller hi there, will this be ever implemented? I love radium and I like it better over others alternatives. I checked styled-components but it is not that good for my own needs.

jaredbrookswhite:
I've made a PR to handle this, #909. Handles multiple animations by passing an array of Keyframes objects.

PR #909, opened 2017-07-13
Handle multiple animationName props

jaredbrookswhite:
Method to handle #675.


src/plugins/keyframes-plugin.js
```diff
@@ -9,11 +9,30 @@ export default function keyframesPlugin(
...
       let value = style[key];
-      if (key === 'animationName' && value && value.__radiumKeyframes) {
-        const keyframesValue = (value: Keyframes);
-        const {animationName, css} = keyframesValue.__process(config.userAgent);
-        addCSS(css);
-        value = animationName;
+      if (
+        key === 'animationName' &&
+        value &&
+        (value.__radiumKeyframes || Array.isArray(value))
```
kylecesmat: Could we do this check above the `if` block?

```
let value = style[key];
const isKeyframeArray = Array.isArray(value);

if (
  key === 'animationName' &&
  value &&
  (value.__radiumKeyframes || isKeyframeArray)
) {
  if (isKeyframeArray) {
   ...etc
  }
```

src/plugins/keyframes-plugin.js
```diff
@@ -9,11 +9,30 @@ export default function keyframesPlugin(
...
+              return animationName;
+            })
+            .join(', ');
+        } else {
+          const keyframesValue = (value: Keyframes);
+          const {animationName, css} = keyframesValue.__process(
+            config.userAgent,
+          );
+          addCSS(css);
+          value = animationName;
```
kylecesmat: can we reduce the duplicate logic here and extract this to a shared method?

ie:

```
const processKeyframeStyle = (value) => {
  const keyframesValue = (value: Keyframes);
  const { animationName, css } = keyframesValue.__process(config.userAgent);
  addCSS(css);
  return animationName;
}
```

```
if (Array.isArray(value)) {
  value = value.map(processKeyframeStyle).join(', ');
} else {
  value = processKeyframeStyle(value);
}
```

kylecesmat:
I believe you need to rebase/merge the upstream `master` branch. Looks like there is a merge conflict with the remote `app.jsx`

jaredbrookswhite:
updated and done. sorry for the confusion, not in my normal dev environment today.

kylecesmat:
This has been released in `radium@0.25.0`

PR #934, opened 2017-10-10
Fix wrong link in Using Radium docs

### LINKED ISSUE #933: Wrong link in "Using Radium" after React updated docs
The link in [Using Radium](http://formidable.com/open-source/radium/docs/guide) docs page  to the React inline styles docs is broken after React updated their docs on 28 September 2017.

"If you're not familiar with handling inline styles in React, see the React guide to the subject **[here](http://facebook.github.io/react/tips/inline-styles.html)**."

Right now the word **here** has a link to http://facebook.github.io/react/tips/inline-styles.html but should point to https://reactjs.org/docs/dom-elements.html#style now that there's not an "Inline Styles" section.

aarongarciah:
Fixes #933

### DIFF
--- a/docs/guides/README.md
+++ b/docs/guides/README.md
@@ -65,7 +65,7 @@ class Button extends React.Component {
 Button = Radium(Button);
 ```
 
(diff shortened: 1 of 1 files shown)


ryan-roemer:
Thanks for the PR @aarongarciah !!!

aarongarciah:
You're welcome 😄

PR #936, opened 2017-10-13
Bug: handle native ES classes.

### LINKED ISSUE #576: Radium wrapping not compatible with native classes?
I've been successfully using Radium with the standard `es2015` Babel preset.

Since our generated code runs on Node 5, I wanted to try out `babel-preset-node5`. With this preset, Babel is no longer translating `class` to ES5 constructs, since Node 5 supports classes out of the box.
However, when wrapping our component classes with Radium, I get this error:

> TypeError: Class constructors cannot be invoked without 'new'

Here is a code snipped which reproduces the error:

``` javascript
import React from 'react';
import radium from 'radium';

class Foo extends React.Component {
  render () {
    return <div>Hello</div>;
  }
}

const WrappedFoo = radium (Foo);

new WrappedFoo (); // TypeError: Class constructors cannot be invoked without 'new'
```

My `.babelrc`:

``` json
{
  "presets": ["stage-0", "node5", "react"]
}
```

There seems to be some logic in the Radium wrapper that is not compatible with
_real_ ES2015 classes.

ryan-roemer:
Here's a stab at getting native ES classes supported in Radium by hooking in and rewriting how babel would _normally_ transpile something like `class Foo extends Bar` and make it use `new` straight up to avoid native ES classes forbidding instantiation with anything besides `new`.

I'd love some folks to test out in "real" projects. Although I've got regression tests here and it works in @onpaws ' https://github.com/onpaws/RadiumErrorDemo repo.

Status

* Should **work now**! I've got a proxy wrapper around a native ES class that makes it invokable _without_ `new` like all other babel-ified classes using an analogous technique that calls real `new` under the hood.

Notes

* Fixes #576
* Adds node tests for just the enhancer to get us at least a little bit of SSR testing
* Improves the babel script commands

@epsitec @begincalendar @onpaws @andrewfrankel / other folks following #576 -- If folks with real applications could try this branch out, it would be a huge validation help! I _think_ this is correct now, but I've only got very basic rendering tests done in the unit test suite. Thanks!

/cc @alexlande


ryan-roemer:
@epsitec @begincalendar @onpaws @andrewfrankel / other folks following #576 -- I've removed the "WIP" status now. I _think_ this is a correct implementation of a native ES class wrapper, but I'd love to have folks pull this branch down and test in real applications...

/cc @alexlande

begincalendar:
Thank you for doing this.

All I have to say so far is that it seems as if this pull request contains at least two disjoint changes:

- The fix for #576 
- Improvements to the build/test setup of the project

Perhaps for future pull requests (for the sake of making them simpler to review), disjoint changes be kept separate.

ryan-roemer:
@begincalendar -- Does the fix work for you?

begincalendar:
Unfortunately I no longer have projects I can user acceptance test the changes against, largely because I moved away from vanilla JS and Babel, towards TypeScript. 

So my assistance on this will be limited to reviewing/critiquing the changes (which I am in the process of doing).

ryan-roemer:
@begincalendar -- Haha, well that's welcome too as we've got some egregious ES/classical constructor hackery going on, so all comments + observations most welcome!

onpaws:
Great to see this get attention!
I've just brought it up to our tester and expect him to be happy :)

ryan-roemer:
Released in `radium@0.19.6`

PR #937, opened 2017-10-18
fixBug: React.createClass is not a function

khaninD:
After React version 15.5 we can't use React.createClass, [https://reactjs.org/blog/2017/04/07/react-v15.5.0.html](url), create-react-class fix it.

### DIFF
--- a/examples/app.jsx
+++ b/examples/app.jsx
@@ -14,7 +14,7 @@
 /* eslint-disable no-use-before-define */
 
 const React = require('react');
-
+const createReactClass = require('create-react-class');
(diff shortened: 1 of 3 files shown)


ryan-roemer:
@khaninD -- Thanks so much for noticing this and reaching out to fix it!

I'm actually going to close this PR in favor of https://github.com/FormidableLabs/radium/pull/939 which just refactors to real `class`es so we don't need to add another dependency and get to modernize our examples. If you get a chance, I'd love for you to try out 939 and make sure that works for you!

PR #940, opened 2017-10-25
Remove trailing commas + rules.

### LINKED ISSUE #932: Infra: Remove prettier trailingComma
And then run `yarn run fixlint`

ryan-roemer:
Trailing commas add no "value" for correctness, and their absence is "always correct" for thing like older version of Node. Now that we actually lint _all_ our JS files in Radium, it is easier to just remove trailing commas in lint than have different lint for transpiled `src` vs. "real node" root `*.js` files like webpack configs, etc.

Fixes #932

/cc @alexlande

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -274,7 +274,6 @@ rules:
       "error",
       {
         "bracketSpacing": false,
-        "trailingComma": "all",
         "singleQuote": true
       }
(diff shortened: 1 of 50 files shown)


alexlande:
👍  🚢  :shipit:

PR #945, opened 2017-11-29
Upgrade to inline-style-prefixer v3

### LINKED ISSUE #918: [TRACKING] Update to use inline-style-prefixer v3
Just as the title says... The newest major version of i-s-p is in use by most other libraries, and we've experienced breakages when used with radium wrapped components that use v2.

exogen:
This upgrades inline-style-prefixer to version 3. Fixes #918.

* There is one failing test due to rofrischmann/inline-style-prefixer#147. If inline-style-prefixer does not incorporate this change eventually, we'll have to adjust its output ourselves. Thoughts?
* Because inline-style-prefixer's [default browser support has changed](https://github.com/rofrischmann/inline-style-prefixer/commit/3e3a7a864399b306be74a0fa05b97e61a81354dd#diff-04c6e90faac2675aa89e2176d2eec7d8L33), and (per @alexlande) Radium roughly targets React's browser support, we now create a custom Prefixer for use with the prefixer plugin. The new Prefixer matches inline-style-prefixer v2's browser support, so people should be able to upgrade without breaking anything.
* The browser data for the new Prefixer is generated by the new `update-prefix-data` script and stored in `src/prefix-data` (the script outputs ES source files).
* New Flow interfaces were added for the new inline-style-prefixer modules we use.


.eslintrc
```diff
@@ -176,7 +176,8 @@ rules:
 
     brace-style:                   # enforce one true brace style
         [2, "1tbs", { "allowSingleLine": true }]
-    camelcase: 2                   # require camel case names
+    camelcase:                     # require camel case names, except in properties
+        [2, { "properties": "never" }]
```
exogen: inline-style-prefixer's browser data generator outputs objects with keys like `ios_saf`. This disables the `camelcase` check for object properties. We can't easily add `eslint-disable` rules to those specific files because they're generated.

ryan-roemer: Which files are generated?

scripts/update-prefix-data.js
```diff
@@ -0,0 +1,27 @@
+/**
+ * Update the files in `src/prefix-data` to match the `browserList` below. Run
+ * this if browser support changes or `inline-style-prefixer` gets fixes for the
+ * supported browsers.
+ */
+import path from 'path'
```
ryan-roemer: We need to lint the scripts.

exogen: Done. I changed ESLint to lint everything so this forgotten-directories thing stops happening (turns out `test` was also not linted).

.eslintrc
exogen: `src/prefix-data/static.js` and `src/prefix-data/dynamic.js`

ryan-roemer: Is there anything else in `prefix-data`? If not, let's can't we just revert this here and drop a new `src/prefix-data/.eslintrc` that overrides it there?

exogen: I considered that and would be happy to do it, lemme make the case for why I didn't though:
* We'd need it in `scripts` too (the input object to `generateData` also has underscores)
* Really I find the whole lint rule applying to properties extremely presumptuous. Other people's APIs use underscores (this example, JSON from REST APIs, even React has `unstable_foo` methods that people actually use) and it's lame needing an exception for that every single time. If our code were completely self-contained and didn't interact with anyone else's APIs then it would make sense.

ryan-roemer: OK.

ryan-roemer:
@exogen -- Looks like https://unpkg.com/inline-style-prefixer@4.0.0/ is released!

Do we just version bump to finish this PR or does other work remain?

/cc @alexlande

test/enhancer-test.js
```diff
@@ -1,37 +1,41 @@
...
-const assertValidEnhancedComponent = (Composed) => {
+const assertValidEnhancedComponent = Composed => {
   const Enhanced = Enhancer(Composed);
   const rendered = render(Enhanced);
 
   expect(render(Composed)).to.not.equal(rendered);
-  expect(rendered)
-    .to.contain('data-radium="true"').and
-    .to.contain('style="background:red;color:white"');
+  expect(rendered).to
```
ryan-roemer: Is this an auto-formatted result?  Chai assertions are typically always:

```js
expect(foo)
  .to.SOMETHING.and
  .to.SOMETHING_ELSE.and
  .to.LAST_THING;
```

ryan-roemer: Or is that prettier turned loose?

stefvhuynh: looks like that is what `prettier` wants. `npm run lint` will fail otherwise.

stefvhuynh:
looks like `inline-style-prefixer@4.0.0` has been released. according to the changelog (https://github.com/rofrischmann/inline-style-prefixer/blob/master/Changelog.md), the only changes are @exogen's PR and updated browser versions. seems like no other work needs to be done here?

ryan-roemer:
@stefvhuynh -- Give the update a go and see if you can get CI to pass!

ryan-roemer:
@alexlande -- Thanks!

@stefvhuynh -- You're a go for merge + publish!

ryan-roemer:
@stefvhuynh @alexlande @exogen -- What semver bump is this change?

exogen:
@ryan-roemer Big dependency bumps like this are tricky – the intention here was to be MINOR (we tried to make sure to support the same browser versions as before, none of our tests had to change, we get some new i-s-p features). But then again i-s-p v3 was also a complete rewrite, it wouldn't surprise me if some edge cases were different for some people, so MAJOR would be the absolute safest.

alexlande:
Given that Radium is still on `0.x` releases I think it would make sense to make this `0.20.0`

PR #947, opened 2017-11-30
Remove decorator recommendations/examples from the docs

### LINKED ISSUE #924: Stop recommending decorators
Let's update the docs to suggest wrapping with Radium the old fashioned way: `Radium()`. The decorators proposal has been going through some changes, now there are two Babel plugins depending on if you want the new or the legacy behavior, and the whole thing is getting a little more complicated.

Plus, functional components have become much more popular since we introduced the decorator syntax, and it's simpler to just refer to one way of applying Radium that will work for any component.

exogen:
This is a simple change that replaces decorators with wrapping everywhere in the docs. Decorators are no longer mentioned at all. Fixes #924.

@alexlande Do we want to at least *mention* decorators somewhere in the docs? (If so, where?)

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,7 +49,7 @@ For a short technical explanation, see [How does Radium work?](#how-does-radium-
 
 ## Usage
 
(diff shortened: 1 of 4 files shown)


alexlande:
I'd sort of prefer to not mention them at all. If it becomes an issue we could add a FAQ item about it, but I don't want to encourage anyone to use outdated syntax.

ryan-roemer:
LGTM!

@alexlande -- any other changes before @exogen merges?

ryan-roemer:
And should we programmatically deprecate the decorator? (Can we? I assume so...)

alexlande:
This LGTM.

> And should we programmatically deprecate the decorator? (Can we? I assume so...)

Don't know if that's necessary, but we could think about it.

Issue #524, opened 2016-01-13
Elements that are hidden when in the hover state come back with the hover state active


basarat:
E.g. if you have some hover styles on a button. And when you click that button the button is hidden. When the button comes back since it never got the `onMouseLeave` it comes back with the hover state active. 
## JSBin

https://jsbin.com/yesisoxiti/1/edit?html,js,output
## Code

``` html
<!DOCTYPE html>
<html>
<head>
<script src="https://cdn.rawgit.com/lodash/lodash/3.0.1/lodash.min.js"></script>
<script src="//fb.me/react-0.13.1.js"></script>
<script src="https://npmcdn.com/radium@latest/dist/radium.js"></script>
  <meta charset="utf-8">
</head>
<body>

</body>
</html>
```

``` jsx
// noprotect

var style = {
   ':hover': {
       background: 'red'
   }
};

@Radium
class App extends React.Component {
  constructor(){
    super();
    this.state = {
      shown : true
    };
  }
  render() {    
    return(  
      <div>

        <button key={1} onClick={()=>this.setState({shown:true})}>
          Show
        </button>    

        { this.state.shown &&
          <button key={2} style={style} onClick={()=>this.setState({shown:false})}>Hide</button>
        }

      </div>
    );
  }          
}

React.render(<App/>, document.body);

```
## Demo

![onhide](https://cloud.githubusercontent.com/assets/874898/12285743/74e7a2a8-ba11-11e5-98bc-6d17ee30113e.gif)

:rose:

ianobermiller:
Thanks for the easy repro! The problems with onMouseLeave are a known issue, but in this specific instance we should be able to clear out the state when an element is no longer rendered. I can't think of a pretty way to do this yet, but it seems doable.

yrokhlin:
I am running in to this issue as well, I was thinking maybe there could be an API call we could make like `radium.clearHover()` that clears all current hover state. It's a workaround, at best, but could be useful.

rcoy-v:
Any updates on this bug? Currently I am having to implement my own hover with state and the associated event handlers.

bcardi:
I just hit this bug today. Any update?

alexlande:
Hey all, no progress on this one to report. Radium might start using generated CSS for pseudo-classes (see https://github.com/FormidableLabs/radium/issues/510), which would resolve this, but that's not in progress either. Ideas and PRs are obviously welcome in the meantime, but apologies for the inconvenience.

[...]

PR #955, opened 2018-01-03
[WIP] clear radium state for unmounted components

stefvhuynh:
Initial attempt to resolve https://github.com/FormidableLabs/radium/issues/524.

1. Keep track of children keys when going through them recursively in `resolve-styles.js`
2. Check these keys against the state in `enhancer.js`
3. Remove the keys in the state that correspond with unmounted components in `enhancer.js`

There is still a lot of cleanup to do, but I'm looking for opinions on the approach.

### DIFF
--- a/examples/app.js
+++ b/examples/app.js
@@ -95,23 +95,35 @@ const VisitedLink = Radium(() => (
   render() {
     return (
       <div>
-        <button onClick={() => this.setState({ showButton: true })}>Show</button>
-        {this.state.showButton && (
-          <button
-            onClick={() => this.setState({ showButton: false })}
-            style={hideButtonStyle}
-          >
-            Hide
-          </button>
-        )}
+        <button key={0} onClick={() => this.setState({ showButton: true })}>Show</button>
+        <div>
+          {this.state.showButton && (
+            <button
+              key={1}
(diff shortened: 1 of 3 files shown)


stefvhuynh:
@ryan-roemer, @ianobermiller, this is an initial attempt at resolving this issue. Can I get thoughts on the approach?

alexlande:
General approach looks solid. I'm a little nervous about the check in `componentDidUpdate` because it could be called extremely frequently in some cases (like if you're doing animations), but I can't think of a better way to handle it.

stefvhuynh:
@alexlande do you think react would batch the state updates if we set it in `componentWillUpdate` instead?

stefvhuynh:
@alexlande, another possible approach would be to expose `setState` on radium as suggested by some people in the issue. though, it seems like it would make sense to clear style states if a component is removed anyway.

alexlande:
@stefvhuynh Yeah, it doesn't feel like something that users should have to handle, if we can do it automatically.

stefvhuynh:
@alexlande, alright, i'll clean this up then.

stefvhuynh:
continued in https://github.com/FormidableLabs/radium/pull/956

PR #956, opened 2018-01-05
clear radium style state in unmounted components

stefvhuynh:
Resolves https://github.com/FormidableLabs/radium/issues/524.

This PR will successfully wipe the interactive `state` (e.g. `:active`, `:focus`, etc) for any child elements that are unmounted so long as a custom key is assigned to the child element.

If there are multiple child components that require interactive styles, `radium` expects a `key` prop to be assigned to each one. In the event where there is only one child with interactive styles, it will autogenerate a key called `main` if no key is provided. Under the hood, every child that does not have a custom `key` will be assigned `main` but the keys are only checked when the `state` is set or retrieved (thus ensuring only the children that have interactive styles are checked). Because of this, it's impossible to know exactly which child is `main` synchronously. 

So for now, a custom `key` prop MUST be provided if removing the interactive style from an unmounted child is important.


stefvhuynh:
@alexlande, @ryan-roemer, cleaned up PR with tests ready for review.

src/__tests__/clean-state-key-test.js
```diff
@@ -0,0 +1,12 @@
+import cleanStateKey from 'clean-state-key.js';
```
ryan-roemer: GLOBAL NIT: Don't need `.js` suffixes here and elsewhere in this PR.

src/__tests__/get-state-test.js
```diff
@@ -1,7 +1,7 @@
 import getState from 'get-state.js';
```
ryan-roemer: GLOBAL NIT: Can you clean up the existing ones with `.js` in imports too? Thanks!

src/__tests__/radium-test.js
```diff
@@ -268,6 +268,46 @@ describe('Radium blackbox tests', () => {
     expect(nav.style.color).to.equal('blue');
   });
 
+  it('resets state for unmounted components, Issue #524', () => {
+    @Radium class TestComponent extends Component {
```
ryan-roemer: Decorator is deprecated, so let's not add _new_ tests with it...

examples/app.js
```diff
@@ -86,6 +86,48 @@ const VisitedLink = Radium(() => (
   </a>
 ));
 
+@Radium class HideButton extends React.Component {
```
ryan-roemer: GLOBAL: Decorator is deprecated so let's not add new code examples with it.

src/__tests__/radium-test.js
```diff
@@ -268,6 +268,46 @@ describe('Radium blackbox tests', () => {
...
+    expect(spans[0].style.color).to.equal('red');
+
+    TestUtils.Simulate.click(spans[0]);
+    spans = getElements(output, 'span');
+    expect(spans.length).to.equal(0);
+
+    TestUtils.Simulate.click(button);
+    spans = getElements(output, 'span');
+    expect(spans.length).equal(1);
+    expect(spans[0].style.color).to.equal('blue');
```
ryan-roemer: NIT: Better written as:

```js
expect(spans)
  .to.have.length(1).and
  .to.have.deep.property('[0].style.color', 'blue);
```

src/clean-state-key.js
```diff
@@ -0,0 +1,7 @@
+/* flow */
+
+const cleanStateKey = key => {
+  return key === null || key === undefined ? 'main' : key.toString();
```
ryan-roemer: NIT: `typeof key === "undefined"` instead of `key === undefined`

src/test-helpers.js
```diff
@@ -16,6 +16,15 @@ export function getElement(output, tagName) {
   );
 }
 
+export function getElements(output, tagName) {
+  return TestUtils.scryRenderedDOMComponentsWithTag(
+    output,
+    tagName
+  ).map(component => {
```
ryan-roemer: NIT:

```js
.map(component => ReactDOM.findDOMNode(component))
```

alexlande:
Code looks great, and works as expected.

> So for now, a custom key prop MUST be provided if removing the interactive style from an unmounted child is important.

Can you add a note in the FAQ about this, and link to it from the "Browser states" section of the guide?

stefvhuynh:
@alexlande, @ryan-roemer, comments addressed. I can merge and release. I did make a change to the `resolveStyles` function so that the output is different than before. I don't know if this will impact the versioning. Thoughts?

ryan-roemer:
@stefvhuynh -- The change to `resolveStyles` is just adding an optional parameter?

stefvhuynh:
@ryan-roemer, it's an optional parameter as well as the output. Instead of just outputting a react element, it outputs an object with the element as one of the values.

Issue #958, opened 2018-01-17
Prefixed inline styles are not resolved


sunny:
Inline styles do not generate valid CSS properties, when using the (default) prefix plugin.

```js
const React = require("react") // v16.2.0
const Radium = require("radium") // v0.19.6
const ReactDOMServer = require("react-dom/server") // v16.2.0

const Foo = () => <div style={{display: "flex"}} />
const StyledFoo = Radium()(Foo)

ReactDOMServer.renderToString(<StyledFoo />)
```

The generated style (tested in [RunKit](https://npm.runkit.com/npm)) gives invalid CSS properties : 

```html
<div style="display:-webkit-box,-moz-box,-ms-flexbox,-webkit-flex,flex" … />
```

When it should be :

```html
<div style="display:-webkit-box;display:-moz-box;display:-ms-flexbox;display:-webkit-flex;display:flex" … />
```

According [to the inline-style-prefixer documentation](https://inline-style-prefixer.js.org/docs/guides/ResolvingArrays.html), array values should be resolved, using something like css-in-js-utils's `resolveArrayValue`, instead of joining the values by `;`.

ryan-roemer:
@sunny -- Would you be able to take a fork of Radium and create a failing unit test? That would be great for helping us identify exactly what is going wrong. Here's the folder where our SSR tests reside: https://github.com/FormidableLabs/radium/blob/master/test/

sunny:
Hey @ryan-roemer ! Sure, here is a failing test. -----^

This seems related to #951, which was resolved by… removing prefixes altogether.

ryan-roemer:
@sunny -- Thanks! We'll hopefully dig in next week!

sunny:
\o/ Thank you for the timely fix 🎉

PR #959, opened 2018-01-17
Prefixed inline styles are not resolved

sunny:
Work-in-progress to have a failing test for issue #958

### DIFF
--- a/src/__tests__/radium-test.js
+++ b/src/__tests__/radium-test.js
@@ -19,6 +19,20 @@ describe('Radium blackbox tests', () => {
     sandbox.restore();
   });
 
+  it('returns only one display attribute', () => {
+    @Radium class TestComponent extends Component {
+      render() {
+        return <div style={{display: 'flex'}} />;
+      }
+    }
+
+    const output = getRenderOutput(<TestComponent />);
+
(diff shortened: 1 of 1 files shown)


src/__tests__/radium-test.js
```diff
@@ -19,6 +19,20 @@ describe('Radium blackbox tests', () => {
...
+    @Radium class TestComponent extends Component {
+      render() {
+        return <div style={{display: 'flex'}} />;
+      }
+    }
+
+    const output = getRenderOutput(<TestComponent />);
+
+    expect(output.props.style).to.deep.equal({
+      display: 'flex'
```
sunny: This returns an array of `['-webkit-box', 'moz-box', '-ms-flexbox', 'webkit-flex', 'flex']`.

sunny: Perhaps this test should try to match `{ display: "-webkit-box;display:moz-box;display:-ms-flexbox;display:webkit-flex;display:flex" }` instead.

ryan-roemer:
Closing in favor of https://github.com/FormidableLabs/radium/pull/962 where we'll do the fix. Thanks for bootstrapping us @sunny !

PR #960, opened 2018-01-17
[BUG] Call componentDidUpdate() super.

### LINKED ISSUE #957: 0.21.0: componentDidUpdate() is not called after changing property in Radium-decorated component
Since Version 0.21.0 I observed the following wrong behaviour of a Radium-decorated React component:

Changing a property from outside of a Radium-decorated component correctly triggers a render() call of this component. But after the render() call there should be a call of componentDidUpdate(). Since 0.21.0 this call is not done. It's still working as expected in 0.19.6 and also without Radium-decoration.

React Version: 16.2.0
Radium Version: 0.21.0

Reproducable by this example (click on output to update data):

```import * as ReactDOM from "react-dom";
import * as React from "react";
import * as Radium from "radium";

class Main extends React.Component {

	constructor(props) {
		super(props);
		this.state = {data: 123};
	}

	render() {
		return <div onClick={() => this.setState({data: Math.random()})}>

			<Test data={this.state.data}/></div>
	}
}

@Radium
class Test extends React.Component {

	componentDidUpdate(prevProps, prevState) {
		//this should be called, but is not
		console.log("componentDidUpdate");
	}

	render() {
		return <div>{this.props.data}</div>
	}
}

ReactDOM.render(
	<Main/>,
	document.getElementById('app')
);
```

ryan-roemer:
Fixes #957 

/cc @q-man @stefvhuynh @kenwheeler

### DIFF
--- a/src/enhancer.js
+++ b/src/enhancer.js
@@ -202,7 +202,11 @@ export default function enhanceWithRadium(
     }
 
     /* eslint-disable react/no-did-update-set-state, no-unused-vars */
-    componentDidUpdate() {
+    componentDidUpdate(prevProps, prevState) {
+      if (super.componentDidUpdate) {
+        super.componentDidUpdate.call(this, prevProps, prevState);
+      }
+
       if (this._extraRadiumStateKeys.length > 0) {
         const trimmedRadiumState = this._extraRadiumStateKeys.reduce(
           (state, key) => {


stefvhuynh:
Lgtm

ryan-roemer:
Released in `radium@0.21.1`

PR #967, opened 2018-02-06
Task: `es-build`

kale-stew:
Closes #966: Output `es/` directory with ESM export/imports intact. 

*Note:* Tests are failing at the moment because of depreciation between Node 4 and `babel-env`.

PR #968, opened 2018-02-07
ESM build, misc other tweaks

### LINKED ISSUE #966: Add `es` build to Radium.
- [x] Output `es/` directory with ESM export/imports intact.
- [x] Add `package.json` fields `module` and `jsnext:main` pointing to `es/index.js`
- [x] Remove `babel-plugin-transform-es2015-modules-commonjs` and switch `es2015` preset to `babel-preset-env`
- [x] gitignore `es`
- [x] whitelist `.npmignore.publishr` adding `!/es` to actually publish to npm
- [x] ... figure out if I've missed anything else... 😛 

Prior examples in other repos:

- https://github.com/FormidableLabs/freactal/pull/94/files (ignore the umd parts)
- https://github.com/FormidableLabs/redux-little-router/pull/247/files
- https://github.com/FormidableLabs/urql/pull/4/files (Note uses babel7/next -- you should use normal babel without `@babel/` prefix)

ryan-roemer:
This builds on @kale-stew 's https://github.com/FormidableLabs/radium/pull/967 and goes further down the rabbit hole.

## Working Items

- `babel-preset-add-module-exports` does not seem to work in producing `module.exports = exports["default"]` in e.g. `lib/index.js`
- Confusingly, the plugin _is_ needed for our frontend Karma test stuff and seems to work there! 😕  And we have mocked stuff like

    ```js
    const resolveStyles = require('inject-loader!resolve-styles')({
      exenv: require('__mocks__/exenv')
    });
    ```
  
  that plays poorly with removing `babel-preset-add-module-exports` (I tried replacing all `require("foo")` with `require("foo").default`) 

## Tasks

@kale-stew --

- [x] Look for all instances of "server" or "SSR" or "Node" or "Node.js", etc. and replace `require('radium')` with `require('radium').default`.
- [x] Places to look: `README.md`, `docs/**`, `examples/**`...
- [x] Look for any remaing `require()`s and convert client code / babel'ed code examples to `import Radium from 'radium';` or other analogous thing.

@ryan-roemer --

- [ ] Verify `dist` works as expected with `global.Radium` variable in prod + dev modes.
- [x] Get `yarn run test-frontend` to pass.
- [ ] Check CI passes...
- [ ] ~~Try to remove all legacy things like `Radium.PROPERTY_NAME =`~~

_Done_

- [x] Figure out if keeping `babel-preset-add-module-exports`
- [x] Figure out if breaking API with removing `module.exports` thing
- [x] Checking if `node4` works in Travis and Appveyor.

## Changes

- Output es/ directory with ESM export/imports intact. Closes #966
- Updates to `babel-preset-env`
- Convert Node.js server example to use `require()`s
- Use `builder concurrent` to parallelize a lot of our build / test stuff
- Remove `{__clearStateForTests, __setTestMode}` as named ESM exports because they're not. Switch to just using `resolveStyles` object directly.
- Remove `package-lock.json` (we'll stick with yarn)

/cc @alexlande @kale-stew


ryan-roemer:
*The Problem*

We _do_ rely on the `babel-preset-add-module-exports` plugin to convert:

```js
export default Radium;
```

to:

```js
exports.default = Radium;
module.exports = exports['default']; // plugin adds this
```

for our legacy compatibility.  However, if we have any other named export, like:

```js

// ESM re-exports
export {Plugins, Style, StyleRoot, getState, keyframes};
```

That I added, the `module.exports = exports['default'];` line is *not* added. This explains why the demo breaks / it's not in babel source but the tests both need the plugin and pass (because they rely on it in other files).

As far as I can tell, only `src/index.js` is affected by this and nothing else apparently has a combination of named + default exports (that are in any way relied on).

docs/api/README.md
```diff
@@ -157,6 +157,12 @@ Allows you to replace the `matchMedia` function that Radium uses. The default is
 
 **Server**
 
+You can require `Radium` on the server / using CommonJS with:
+
+```jsx
+var Radium = require('radium');
```
kale-stew: Do we not need `.default` on this import?

ryan-roemer: Correct. That is the change with the new root `index.js` file.

kale-stew: Got it. 👍

ryan-roemer:
@alexlande @exogen -- I've got this branch and a separate experiment repo for all of the consumption scenarios I could think of. Can you take a look at that (and this PR) and let me know if we're missing anything or anything needs changing for this approach / PR?

https://github.com/FormidableLabs/radium-experiments-v0.22

Thanks!!!

alexlande:
Went through the experiments, everything looks good as far as I can tell. Thanks for all of the work here, module system interop is so complicated 👍

ryan-roemer:
Released in `radium@0.22.0` 🎉

PR #977, opened 2018-03-10
Support ES7 arrow functions on React classes.

### LINKED ISSUE #738: Bug: Radium wrapper breaks on ES7 arrow methods
The current version of Babel is not supporting decorators until the change in decorators is finalized with  TC-39. Because of this I've done this in react:

``` javascript
class LandingHero extends React.Component {
<code here>
}
export default Radium(LandingHero);
```

This works fine, unless render is an arrow method like so:

``` javascript
class LandingHero extends React.Component {
  render = () => <div>Something</div>;
}
export default Radium(LandingHero);
```

Then I get the following error:

``` error
Uncaught TypeError: Cannot call a class as a function
```

Of course the error is coming from the babel runtime compiler and not radium specifically but I was wondering if there was something that could be done here to support this in the future.

ryan-roemer:
- Fixes #738
- This is hacky, but possibly **acceptably** hacky.
- Detect and transfer instance method properties from arrow functions to a copy of the prototype (where by definition they don't exist) to allow property inheritance call chaining by `RadiumEnhancer` which otherwise doesn't get its methods called.

/cc @alexlande @nbrady-techempower @jimsalabim

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Radium Changelog
 
+## Unreleased
+- Support ES7 arrow functions for React class methods. (#738)
+
 ## 0.22.1 (March 1, 2018)
 - Fix `keyframes` bug from prefixed inline styles. (#973)
 
--- a/src/__tests__/radium-test.js
+++ b/src/__tests__/radium-test.js
@@ -742,10 +742,14 @@ describe('Radium blackbox tests', () => {
       };
     }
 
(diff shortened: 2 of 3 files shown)


nbrady-techempower:
@ryan-roemer Thanks for your continued work on this!

jimsalabim:
@ryan-roemer Great work!

ryan-roemer:
Released in `radium@0.23.0`

Issue #950, opened 2017-12-12
Bug: Cannot read property 'children' with rendered children array.


or2008:
I am using the new React feature - portals, to render a div outside of the app root element, it works great, however, when wrapping the component with Radium it throws this error:

Uncaught TypeError: Cannot read property 'children' of undefined
    at resolveStyles (resolve-styles.js:317)

![image](https://user-images.githubusercontent.com/3414979/33881550-c47ba51e-df3d-11e7-95e7-dee0dac3091a.png)

ryan-roemer:
@or2008 -- Can you provide a minimal repository with install + error reproduction steps so that we can more efficiently dive in and diagnose? Thanks!

alex00x:
Hi @ryan-roemer  The error is reproduced in case the render method returns an array

Example:
```
class Spinner extends React.Component {
  render() {
    return [
      <Style
        rules={{
          '.smthng': {
            color: 'green'
          }
        }}
      />,
      <div>
        <div
          style={[spinnerStyles.inner, {'@media print': {height: '10px'}}]}
        />
      </div>
    ];
  }
}

Spinner = Radium(Spinner);
```

ryan-roemer:
Verified. WIP branch up at https://github.com/FormidableLabs/radium/compare/bug/render-array with just failing regression test for a start.

@alex00x -- Can you review and confirm that failing test captures the bug?

alex00x:
@ryan-roemer Yep, looks like

kylecesmat:
This bug was fixed & released in `Radium@0.24.0`. If you are still having issues, please re-open but I will close this as resolved :)

PR #979, opened 2018-03-28
Bug: Render array

chrisbolin:
React 16 introduced [rendering arrays of elements](https://reactjs.org/blog/2017/09/26/react-v16.0.html).

```jsx
render() {
  return [
    <li key="A">First item</li>,
    <li key="B">Second item</li>,
    <li key="C">Third item</li>,
  ];
}
```

This usage threw an error in Radium: https://github.com/FormidableLabs/radium/issues/950

### DIFF
--- a/src/resolve-styles.js
+++ b/src/resolve-styles.js
@@ -385,6 +385,15 @@ resolveStyles = function(
     );
   }
 
+  // TODO HERE
+  // renderedElement can be an array, not a single element.
+  //
+  // https://github.com/FormidableLabs/radium/issues/950
+  //
+  if (Array.isArray(renderedElement) && !renderedElement.props) {
+    console.log("TODO HERE -- ARRAY WITHOUT PROPS");
+  }
+
   // ReactElement
   if (
     !renderedElement ||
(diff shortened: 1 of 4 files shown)


chrisbolin:
Note that Radium requires the use of keys. From the [Guide](http://formidable.com/open-source/radium/docs/guide)...

> Radium allows you to style multiple elements in the same component. You just have to give each element that has browser state modifiers like :hover or media queries a unique `key` or `ref` attribute.

That requirement still holds true with rendered arrays. (In fact, _React_ requires keys for rendered arrays)

chrisbolin:
The next step (can be in a different PR) will be testing `<Fragment>`, which was introduced in [React 16.2](https://reactjs.org/blog/2017/11/28/react-v16.2.0-fragment-support.html). We'll need to update `yarn.lock` to `16.2.0`

src/resolve-styles.js
```diff
@@ -385,6 +385,34 @@ resolveStyles = function(
     );
   }
 
+  // TODO HERE
```
ryan-roemer: Let's clean up all `TODO` notes 😉

chrisbolin:
Merging. Will tackle testing `<Fragment>` rendering next (and fix things if necessary)

PR #989, opened 2018-05-11
Make <StyleSheet> manually update <style> tag DOM

mhink:
Fixes #626

PR #994, opened 2018-07-06
Fix animation race condition on Safari

apiv:
Fixes #192 :

On Safari, there is a race condition, in which the `@keyframes` are added to the `<StyleSheet>` on the page after the `animationName` style property is added to the styled element. Safari doesn't find the animation, and doesn't refresh those elements when the animation is added to the stylesheet.

Looking at the code, the update of the global stylesheet (`src/components/style-sheet.js:43`) is deferred by 0 milliseconds, while the addition of the style rule to the element is synchronous. This would cause the race condition described above.

This PR removes the debouncing from the CSS change listener, but leaves the debounce in `componentDidMount` intact.


apiv:
Closing because the same problem is solved by #989

Issue #999, opened 2018-09-04
Weird problem with meteor 1.7 / es6 classes and binding of this in constructor


dobesv:
When trying to upgrade our app to meteor 1.7 we get some really weird behavior.

It seems like the value of `this` in the constructor is not the same `this` that is used later on in `render`.  As a result, if you have bound your methods to `this` using arrow syntax or `this.foo = this.foo.bind(this)`, it only has access to the props at component creation time.

I have created a reproduction over here in the `react` branch of my fork of `meteor/todos`: https://github.com/dobesv/todos/tree/react

Note that in https://github.com/dobesv/todos/blob/react/imports/ui/components/TodoItem.jsx I have added some boolean strings to the display, indicating whether the values of `this.props.editing` are the same between the bound method and the render method.

If I remove `Radium` from that component, the problem goes away.

dobesv:
Looking at the code, I think it might be this rather evil piece of code here:

```
  if (isNativeClass(ComposedComponent)) {
    // Manually approximate babel's class transpilation, but _with_ a real `new` call.
    ComposedComponent = (function(OrigComponent): constructor {
      function NewComponent() {
        // Ordinarily, babel would produce something like:
        //
        // ```
        // return _possibleConstructorReturn(this, OrigComponent.apply(this, arguments));
        // ```
        //
        // Instead, we just call `new` directly without the `_possibleConstructorReturn` wrapper.
        const source = new OrigComponent(...arguments);

        // Then we manually update context with properties.
        copyProperties(source, this);

        return this;
      }

      inherits(NewComponent, OrigComponent);

      return NewComponent;
    })(ComposedComponent);
  }
```

This seems to be copying properties after running the constructor, which assumes that there are no references to the original object inside the properties of the object.

dobesv:
I'm not totally sure what the reasoning behind that code was, but I wonder if it should have preserved the `this` binding on any methods its copies, or perhaps just any react lifecycle methods.

dobesv:
The reason this used to work in meteor 1.6 is that they used to transpile classes with babel, but now they emit the class syntax as-is for the runtime environment (e.g. the browser) to handle.  So Radium now sees a "native" class and that sketchy logic is used.

ryan-roemer:
Here's the background for why we do the native class detection (it's because we use inheritance and babel inheritance plays very poorly with a target native ES class in Node / supported browser platforms): https://github.com/FormidableLabs/radium/issues/576

[...]

PR #1001, opened 2018-09-05
Better es6 class support

dobesv:
Use Reflect to construct class in a more compatible manner.

Reflect should be available in environments that use es6 classes.

This can fix https://github.com/FormidableLabs/radium/issues/999 because we preserve the same value of `this` as was used in the object constructor, whereas the previous implementation discarded that object and used a new object after calling the constructor.  That resulted in some confusing behavior if the constructor stored referenced to `this` or bound methods / functions to `this`.


ryan-roemer:
Thanks for opening this!!!

In case it helps as we're getting ci fixed up on this, probably about time to switch Travis versions to:

```yml
node_js:
  - "6"
  - "8"
  - "10"
```

(Appveyor is already there...)

ryan-roemer:
Something we've added is failing on uglify

```
ERROR in radium.min.js from UglifyJs
Unexpected token: punc (.) [radium.min.js:508,11]
[builder:proc:end:2] Command: webpack --config=webpack.config.minified.js
[builder:proc:error] Code: 2, Command: webpack --config=webpack.config.minified.js
[builder:builder-core:end:5230] Task: concurrent build-dist-dev, build-dist-prod, Error: Command failed: sh -c webpack --config=webpack.config.minified.js
[builder:proc:end:2] Command: builder concurrent --buffer build-dist-dev build-dist-prod
[builder:proc:error] Code: 2, Command: builder concurrent --buffer build-dist-dev build-dist-prod
[builder:builder-core:end:5215] Task: concurrent build-lib, build-dist, build-es, Error: Command failed: sh -c builder concurrent --buffer build-dist-dev build-dist-prod
error Command failed with exit code 2.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

Probably worth looking at transpiled output (webpack should be consuming `es/*` files). If it's an ES6 thing that _should_ be supported, we can look to updating the webpack build to use something more es-friendly...

ryan-roemer:
And to save you round trips to CI, all of this should work/fail the same in localhost:

```sh
$ yarn run build-examples
$ yarn run lint
$ yarn run build
$ yarn run test-coverage
```

dobesv:
Yeah, I'm working on the CI fixes.  I think I don't need to use `new.target`, which is what uglifyjs is mad about.  `this.constructor` should always be available and correct from my understanding.

dobesv:
Looks like tests are passing.  Not sure if you have any other way to verify this won't break things for people who were using es native classes out there.  It does seem to fix the issue for me if I hand-patch node_modules using the generated file from `lib`.

ryan-roemer:
@dobesv --  Great work! You changes still pass the original regression test I wrote for ES native classes at: https://github.com/FormidableLabs/radium/blob/0f7521fd1e5351ed203a5800a1c0a04bec0b9643/test/enhancer-test.js#L56-L66 so I _think_ we should be good.

@boygirl @alexlande -- We've got some great work here that I don't have time right now to kick the tires on. Can you recruit someone to vet this and possibly merge + release? Thanks!!!

dobesv:
Any ETA on this one?

ryan-roemer:
I believe @kylecesmat is going to look at this when he has a chance, hopefully this upcoming week. Thanks for your patience!

(I'd jump in and try to wrangle it, but I'm on parental leave with a newborn 👶 )

test/enhancer-test.js
```diff
@@ -64,4 +64,66 @@ describe('Enhancer', () => {
...
+        ]);
+      }
+    }
+
+    const cleanup = require('jsdom-global')();
+    try {
+      const Enhanced = Enhancer(TestComponent);
+      const elt = React.createElement(Enhanced, {flag: true});
+      const {configure, mount} = require('enzyme');
+      const Adapter = require('enzyme-adapter-react-16');
```
kylecesmat: I do want to see if we can refactor some of our existing tests to use jsdom/enzyme - in which case moving these to top-level requires. But this implementation is great for now.

.travis.yml
```diff
@@ -6,9 +6,9 @@ addons:
   chrome: stable
 
 node_js:
-  - "4"
   - "6"
   - "8"
+  - "10"
```
kylecesmat: Thanks for updating this!

src/enhancer.js
```diff
@@ -92,27 +66,26 @@ export default function enhanceWithRadium(
...
-        //
-        // Instead, we just call `new` directly without the `_possibleConstructorReturn` wrapper.
-        const source = new OrigComponent(...arguments);
-
-        // Then we manually update context with properties.
-        copyProperties(source, this);
-
-        return this;
+        // Use Reflect.construct to simulate 'new'
+        const obj = Reflect.construct(
```
kylecesmat: What are the implications of using `Reflect` in regards to cross-browser support? I believe IE11 is the only browser that will need a polyfill, would you be able to add a note in the README about including a polyfill like https://github.com/zloirock/core-js#ecmascript-reflect?

A note in this section would probably be best - https://github.com/FormidableLabs/radium#importing-radium

kylecesmat:
This has been released in `radium@0.25.0`

src/enhancer.js
dobesv: @kylecesmat I don't think you can create a "native class" in IE11, so this code would never run in that browser.

test/enhancer-test.js
dobesv: Yeah, since I'm not a core contributor I wanted to have minimal impact.

.travis.yml
dobesv: Well, I had to  in order to get a green light on the PR :)

PR #1006, opened 2018-10-09
Apply prettier formatting on codebase

stevenmusumeche:
This PR upgrades prettier and adds a new `npm run format` command which runs it on the entire codebase.  The rules are configured in `.prettierrc`, but I copied them from the `eslint` configuration to minimize the number of changes.


ryan-roemer:
I'm not reviewing as this is @kylecesmat 's game here, but in spectacle, we do our scripts this way:

```js
// package.json
    "prettier": "prettier \"**/*.{js,json,css,md}\"",
    "prettier-fix": "npm run prettier -- --write",
    "prettier-check": "npm run prettier -- --list-different",
```

and run `prettier-check` in CI. It exits non-zero (failing the build) if prettier would change any files.

An alternative / additional way to do things is add the prettier plugins + rules to `eslintrc`.

All good approaches! Hope the tip helps :)

Issue #969, opened 2018-02-08
Tests/Infra: Update babel, webpack, karma stuff.


ryan-roemer:
- [ ] https://github.com/istanbuljs/babel-plugin-istanbul and remove `isparta-loader`
- [ ] Modern webpack
- [ ] Update karma
- [ ] Anything else appropriate

PR #1011, opened 2018-12-05
Update Infra #969 - Replace isparta with istanbul

blainekasten:
Part of #969 
Changes from using `isparta` to `istanbul` for test coverage.


.babelrc
```diff
@@ -8,14 +8,15 @@
     "react"
   ],
   "env": {
-    "commonjs": {
+    "test": {
```
ryan-roemer: We should have a `commonjs` env because that's one of our target `es` (ESM by default) vs. `lib` (CommonJS).

And it should **not** be instrumented. In this branch, running `yarn build-lib` and looking at `lib/**/*.js` now shows all our *library* code that we ship (e.g., https://unpkg.com/radium@0.25.0/lib/) is fully instrumented with a whole bunch of istanbul cruft.

Running `yarn build-lib` builds to `lib` now

ryan-roemer: The `test` environment _could_ be added here, but `build-lib` needs to produce non-instrumented code. Which is why it was added in karma conf babel-loader before.

CONTRIBUTING.md
```diff
@@ -135,3 +135,46 @@ available at [http://contributor-covenant.org/version/1/4][version]
 
 [homepage]: http://contributor-covenant.org
 [version]: http://contributor-covenant.org/version/1/4/
+[version]: http://contributor-covenant.org/version/1/4/
+
+### Architecture
```
ryan-roemer: Like it!

blainekasten: lol, i'm confused. I didn't write any of this..

.babelrc
blainekasten: If we added an additional env `test` along side the old unchanged `commonjs`, how would that affect the output of `build-lib` then?

PR #1012, opened 2018-12-05
Infra/prettier

blainekasten:
Built to supersede #1006 
- Adds prettier
- ~Adds husky/lint-staged to run prettier over the committed files~
- Adds a prettier check to travis to ensure files are formatted
- Runs prettier over the codebase

All of the important changes happen in the [package.json](https://github.com/FormidableLabs/radium/pull/1012/files#diff-b9cfc7f2cdf78a7f4b91a753d10865a2)


ryan-roemer:
I pushed this change (https://github.com/FormidableLabs/radium/pull/1012/commits/56959115ff0ce7a8b2052f29775370f3db0ccf56) to ignore the examples/dist directory. If you build examples _first_, then do format-check it fails on the examples/dist files.

I also added a parser to eslint config to avoid repeated error of "can't find parser" thing.

package.json
```diff
@@ -34,6 +34,8 @@
     "examples": "webpack-dev-server --config examples/webpack.config.js --no-info --content-base examples",
     "examples-server": "babel-node examples/server.js",
     "flow": "flow check",
+    "format": "prettier --write \"{src,examples}/**/*.js\"",
```
ryan-roemer: I experimented with this, but I couldn't see any different from this vs. `yarn fixlint` -- both apply prettier rules to all the same files. Maybe we can remove this one since the lint one is more specific?

package.json
```diff
@@ -34,6 +34,8 @@
...
+    "format-check": "prettier --list-different \"{src,examples}/**/*.js\"",
```
ryan-roemer: I don't think we necessarily need this one easier since eslint will apply **all** the prettier rules to our JS files.

Typically, you see a prettier dirty check when there's something _not_ controlled by eslint prettified like https://github.com/FormidableLabs/spectacle/blob/master/package.json#L27-L29

```js
    "prettier": "prettier \"**/*.{js,json,css,md}\"",
    "prettier-fix": "npm run prettier -- --write",
    "prettier-check": "npm run prettier -- --list-different",
```

Maybe remove this one too and from CI since we're already checking?

package.json
blainekasten: Good catch! I agree we can ditch this.

.eslintrc
```diff
@@ -276,6 +276,7 @@ rules:
       "error",
       {
         "bracketSpacing": false,
-        "singleQuote": true
+        "singleQuote": true,
+        "parser": "babylon"
```
kylecesmat: Just so I'm clear, this prettierConfig will only run against `*.js` files, correct? I do not believe this `parser` flag is necessary anymore as latest versions of prettier can infer what type of parser to use based on file extension.

I have hit issues with this flag specifically in the past when prettier attempts to format `html` or `json` and we force it into using the `babylon` parser.

examples/app.js
```diff
@@ -30,9 +30,9 @@ class HoverMessage extends React.Component {
         <button key="button" style={{display: 'flex', ':hover': {}}}>
           Hover me!
         </button>
-        {getState(this.state, 'button', ':hover')
-          ? <span>{' '}Hovering!</span>
-          : null}
+        {getState(this.state, 'button', ':hover') ? (
+          <span> Hovering!</span>
```
kylecesmat: 🎉

.eslintrc
blainekasten: I'm not 100% sure, but i'm guessing since eslint only runs against js files, that this should be true.

ryan-roemer: @kylecesmat -- I added this here. eslint was producing errors for prettier rule of unspecified parser. And as this is the _eslint_ config and not the prettierc, I'm hoping it should be 🆗

PR #1013, opened 2018-12-08
Fix snapshot argument for componentDidMount - Issue #985

### LINKED ISSUE #985: Enhancer.js doesn't pass snapshot argument in componentDidUpdate
In last React version (16.3.0), which is supported in current package.json, `componentDidUpdate` takes `snapshot` as third argument ([docs](https://reactjs.org/docs/react-component.html#componentdidupdate)). The current implementation of enhancer.js [doesn't consider it](https://github.com/FormidableLabs/radium/blob/master/src/enhancer.js#L254):
```jsx
    componentDidUpdate(prevProps, prevState) {
      if (super.componentDidUpdate) {
        super.componentDidUpdate.call(this, prevProps, prevState);
      } 
```

A potential fix looks simple. Would you be intested in a PR?

blainekasten:
Fixes #985 
-

### DIFF
--- a/src/__tests__/radium-test.js
+++ b/src/__tests__/radium-test.js
@@ -276,6 +276,36 @@ describe('Radium blackbox tests', () => {
     expect(nav.style.color).to.equal('blue');
   });
 
+  it('passes snapshot to the componentDidUpdate of the component, Issue #985', () => {
+    const SNAPSHOT = 'SNAPSHOT';
+
+    class SnapshotComp extends Component {
+      componentDidMount() {
+        this.forceUpdate();
+      }
+
+      componentDidUpdate(props, state, snapshot) {
+        expect(snapshot).to.equal(SNAPSHOT);
+      }
+
+      getSnapshotBeforeUpdate() {
+        return SNAPSHOT;
+      }
+
+      render() {
+        return null;
+      }
+    }
+
+    sinon.spy(SnapshotComp.prototype, 'componentDidUpdate');
+
+    const TestComponent = Radium(SnapshotComp);
+
+    TestUtils.renderIntoDocument(<TestComponent />);
+
+    expect(SnapshotComp.prototype.componentDidUpdate).to.have.been.calledOnce;
+  });
+
   it('resets state for unmounted components, Issue #524', () => {
     class TestComponent extends Component {
       state = {showSpan: true};
--- a/package.json
+++ b/package.json
@@ -106,9 +106,9 @@
     "nodemon": "^1.11.0",
     "object-assign": "^4.1.1",
     "prettier": "^1.15.3",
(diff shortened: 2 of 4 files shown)


src/__tests__/radium-test.js
```diff
@@ -276,6 +276,40 @@ describe('Radium blackbox tests', () => {
...
+  it('passes snapshot to the componentDidUpdate of the component, Issue #985', () => {
+    const SNAPSHOT = 'SNAPSHOT';
+
+    @Radium
+    class SnapshotComp extends Component {
+      state = {
+        incrementToTriggerLifecycles: 0
+      };
+
+      componentDidUpdate(props, state, snapshot) {
```
ryan-roemer: Should we also assert this was actually called just to cover our bases?

blainekasten: Updated, and refactored a bit to be simpler.

src/__tests__/radium-test.js
```diff
@@ -276,6 +276,36 @@ describe('Radium blackbox tests', () => {
...
+      }
+    }
+
+    sinon.spy(SnapshotComp.prototype, 'componentDidUpdate');
+
+    const TestComponent = Radium(SnapshotComp);
+
+    TestUtils.renderIntoDocument(<TestComponent />);
+
+    assert(SnapshotComp.prototype.componentDidUpdate.calledOnce);
```
ryan-roemer: Let's remove `assert` and replace with `sinon-chai` asserts which we use everywhere else.

So, something like:

```js
expect(SnapshotComp.prototype.componentDidUpdate).to.have.been.calledOnce;
// OR
expect(SnapshotComp.prototype.componentDidUpdate).to.have.callCount(1);
```

.eslintrc
```diff
@@ -17,6 +17,7 @@ globals:
     require: true
     module: true
     expect: true
+    assert: true
```
ryan-roemer: I don't think we need `assert` as the new place it was used is probably better served with a first-class `sinon-chai` assert :)

ryan-roemer:
Weird other observation -- coverage seemed to have went way down... I haven't dug in but is there something up there that you can see?

https://github.com/FormidableLabs/radium/pull/1013#issuecomment-445480453

> Coverage decreased (-53.6%) to 41.667%

src/__tests__/radium-test.js
blainekasten: done.

.eslintrc
blainekasten: done.

blainekasten:
@ryan-roemer I'm hoping it's because I pushed in an `it.only` on accident. We'll find out soon! otherwise i have no idea how it would jump that much.

PR #1015, opened 2018-12-20
Fix media query ordering

### LINKED ISSUE #975: Media Queries Out Of Order
screen shot 2018-03-01 at 8 16 30 am
screen shot 2018-03-01 at 8 16 24 am

Most of the time this isn't an issue but I found that the media queries are getting created out of order. Is this something I am doing wrong or is this possibly something with the library?

rawrmonstar:
Fixes #975 

Identical media queries with identical rulesets across different components were being deduped which
broke user-defined media query ordering.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -151,7 +151,7 @@ rules:
     no-undefined: 0                # disallow use of undefined variable
     no-unused-vars: 2              # disallow declaration of variables that are not used in
                                    #    the code
-    no-use-before-define: 2        # disallow use of variables before they are defined
+    no-use-before-define: 1        # disallow use of variables before they are defined
     no-var: 2
 
     ###########################################################################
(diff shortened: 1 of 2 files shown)


.eslintrc
```diff
@@ -150,7 +151,7 @@ rules:
     no-undefined: 0                # disallow use of undefined variable
     no-unused-vars: 2              # disallow declaration of variables that are not used in
                                    #    the code
-    no-use-before-define: 2        # disallow use of variables before they are defined
+    no-use-before-define: 1        # disallow use of variables before they are defined
```
rawrmonstar: I think adding the es6 env made it so eslint started checking this rule... there are 2 instances in the code base from before this PR that don't follow this rule.

rawrmonstar:
Closing this in favor of the old https://github.com/FormidableLabs/radium/pull/635. My solution was basically a less informed version of the work done in that ongoing PR. I didn't know that that PR existed before.

PR #1020, opened 2019-03-20
Update unitless CSS property list and licensing.

bgeihsgt:
This commit solves two issues in append-px-if-needed:

1. The list of properties is out of date for unitless CSS properties.
2. Updates the licensing for this copied code snippet to the more friendly MIT license.


bgeihsgt:
Hoping to get this merged & released within the next week. Would that be possible?

Issue #1034, opened 2019-12-04
:active style getting stuck - mouse_up_listener error showing in console


dancherb:
The style given for ":active" persists, even after releasing the mouse. The console shows the below error. 

```
Uncaught TypeError: Cannot convert undefined or null to object
    at Function.keys (<anonymous>)
    at resolve-interaction-styles-plugin.js:85
    at mouse-up-listener.js:6
    at Array.forEach (<anonymous>)
    at _handleMouseUp (mouse-up-listener.js:5)
```

Demo component:
```

import React, { Component } from 'react';
import Radium from 'radium';

class Checkbox extends Component {
    render() {
        return(
            <div 
                style={{
                    backgroundColor: "red",
                    height: 30,
                    width: 30,
                    opacity: 1,
                    ":active": {
                        opacity: 0.6,
                    }
                }}
            />
        )
    }
}

export default Radium(Checkbox)
```

Versions:

```
"react": "^16.12.0",
"radium": "^0.26.0",
```

andres-torres-marroquin:
I tried that Demo component, I wasn't able to make it fail, however when I converted it to a functional component I the error thereafter.

nathanmarks:
Taking a look now -- we've encountered the same issue

PR #1041, opened 2020-01-22
Fix initial enhancer state for fn components

nathanmarks:
This addresses https://github.com/FormidableLabs/radium/issues/1034

The listener in the plugin where the error is being raised will run and if `_radiumStyleState: {}` hasn't been set by another action yet we have issues. There is a test enforcing this initial state for class components and I have added one for function components and fixed the issue.

cc @dancherb

### DIFF
--- a/src/__tests__/enhancer-test.js
+++ b/src/__tests__/enhancer-test.js
@@ -21,6 +21,24 @@ describe('Enhancer', () => {
     expect(ref.current.state).to.deep.equal({_radiumStyleState: {}});
   });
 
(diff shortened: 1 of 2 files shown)


nathanmarks:
@boygirl just a friendly ping -- can you let me know if we'll be able to merge and release a patch version for this?

kylecesmat (approved):
I do apologize for leaving this PR unmerged for quite a few months - Thank you so much for this fix! It will be released in a *patch* version

